PR #166, opened 2022-07-22
Remove _FILTERED

nikolas-burkoff:
Part of https://github.com/insightsengineering/teal.slice/issues/76

### DIFF
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -248,7 +248,6 @@ srv_g_ae_oview <- function(id,
 
     observeEvent(input$arm_var, {
       ANL <- datasets$get_data(dataname, filtered = FALSE) # nolint
-      ANL_FILTERED <- datasets$get_data(dataname, filtered = TRUE) # nolint
 
       req(!is.null(input$arm_var))
       arm_var <- input$arm_var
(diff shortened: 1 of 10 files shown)


R/tm_g_ae_oview.R
```diff
@@ -288,37 +287,37 @@ srv_g_ae_oview <- function(id,
...
+      assign(anl_name, ANL)
 
       teal.code::chunks_reset(envir = environment())
 
-      validate(need(nlevels(ANL_FILTERED[[input$arm_var]]) > 1, "Arm needs to have at least 2 levels"))
-      validate_has_data(ANL_FILTERED, min_nrow = 10)
-      if (all(c(input$arm_trt, input$arm_ref) %in% ANL[[input$arm_var]])) {
+      validate(need(nlevels(ANL[[input$arm_var]]) > 1, "Arm needs to have at least 2 levels"))
+      validate_has_data(ANL, min_nrow = 10)
+      if (all(c(input$arm_trt, input$arm_ref) %in% ANL_UNFILTERED[[input$arm_var]])) {
```
gogonzo, line 301: ANL_FILTERED is assigned but the condition in above `if`doesn't make sense when I look at the message. 
- it checks if levels are in unfiltered data
- if not, throw message to suggest that data might be filtered out.
In the line 313 there is check again, but on the filtered data saying that "Plot loading". In my opinion we should take 302-312 out out this if and not use ADSL_UNFILTERED at all. Similar validate is in R/tm_g_ae_sub.R but not ADSL_UNFILTERED there


```suggestion
      if (all(c(input$arm_trt, input$arm_ref) %in% ANL_UNFILTERED[[input$arm_var]])) {
```

nikolas-burkoff: we've decided to leave this as is and attack it during https://github.com/insightsengineering/teal.osprey/issues/145

gogonzo (approved):
👍

Issue #13, opened 2021-08-05
Spiderplot facet variables should be distinct


cicdguy:
Should you be allowed to select the same variable to facet for both rows and columns?
Or should we add an extra validate here?
![image](https://user-images.githubusercontent.com/12943682/128376813-5710ecc7-03da-41d7-97a6-8ddc26cc2a06.png)

Provenance: 
```
Creator: nikolas-burkoff
```

kumamiao:
Current selection does not make much sense here, extra validate would be nice to have, thanks!

Melkiades:
Should this be closed as it is in "done"?

PR #197, opened 2022-11-25
x and y facets different validation

BLAZEWIM:
# Pull Request

Fixes #[13](https://github.com/insightsengineering/teal.osprey/issues/13)

### DIFF
--- a/R/tm_g_spiderplot.R
+++ b/R/tm_g_spiderplot.R
@@ -229,6 +229,18 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label, 
   moduleServer(id, function(input, output, session) {
     vals <- reactiveValues(spiderplot = NULL) # nolint
 
+    iv <- shinyvalidate::InputValidator$new()
+    iv$add_rule("paramcd", shinyvalidate::sv_required())
+    iv$add_rule("x_var", shinyvalidate::sv_required())
+    iv$add_rule("y_var", shinyvalidate::sv_required())
+    iv$add_rule("marker_var", shinyvalidate::sv_required())
+    iv$add_rule("line_colorby_var", shinyvalidate::sv_required())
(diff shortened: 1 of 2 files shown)


R/tm_g_spiderplot.R
```diff
@@ -258,6 +258,14 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
...
       validate(need(nrow(ADSL) > 0, "ADSL data has zero rows"))
       validate(need(nrow(ADTR) > 0, paste(dataname, "data has zero rows")))
+      if (length(xfacet_var) * length(yfacet_var) > 0) {
+        validate(
+          need(
+            !anyDuplicated(c(xfacet_var, yfacet_var)),
+            "X- and Y-facet variables must not be duplicated."
+          )
+        )
+      }
```
mhallal1, line 268: is using `shinyvalidate` possible here?

chlebowa:
Is it just me or is `shinyvalidate` quite bloaty?

BLAZEWIM:
> Is it just me or is `shinyvalidate` quite bloaty?

Agree in many cases. You need to initialize class, enable it and then use validate anyway to trigger (+3 lines), also in this example you need to create another class inside the reactive. In many cases syntax can became quite unfriendly. From my perspective the only gain is putting warnings next to the fields.

chlebowa:
> > Is it just me or is `shinyvalidate` quite bloaty?
> 
> Agree in many cases. You need to initialize class, enable it and then use validate anyway to trigger (+3 lines), also in this example you need to create another class inside the reactive. In many cases syntax can became quite unfriendly. From my perspective the only gain is putting warnings next to the fields.

Yeah, the fact that you have to build validators in reactives is additionally confusing. You can't even build them all in one place and then enable as necessary.

I do agree that highlighting the inputs is a good thing, ultimately, I'm just wondering whether there is a less verbose way of achieving that.

chlebowa (approved):

Issue #185, opened 2022-10-07
shiny validate improvements


nikolas-burkoff:
From https://github.com/insightsengineering/teal.osprey/issues/183

We should give better error messages in i.e. shinyvalidate::sv_required() takes a message argument so we can have:

![image](https://user-images.githubusercontent.com/15201933/194595651-5dba761d-1cac-4456-bdf7-295eaccf5dbf.png)

And the advantage of this is that the helpful error message is then available to be displayed in the main output so you can see what's wrong directly rather than only being told "please observe red flags in the encoding" 

![image](https://user-images.githubusercontent.com/15201933/194595869-92f32a2c-5537-4852-a57b-6a2bfaa6fc28.png)

We won't have capacity to do this before the release @donyunardi but it should be considered before we roll shinyvalidate out to other packages

donyunardi:
Thanks Nik.
I added this issue to shinyvalidate roadmap.

chlebowa:
Here's an idea:
`shinyvalidate` messages are collated in the main output and preceded by an optional generic message.

![Image](https://user-images.githubusercontent.com/114988527/203970102-50dab32d-0902-4fcf-8f23-60128b1ba09a.png)


This is done by a function that returns either NULL or the messages, so one would call
```
shiny::validate(need(foo(), message = foo()))
```
This function would be placed in `teal`.

Now, there are more `shinyvalidate::InputValidator` objects in this particular module and currently I am only capturing one, but I would aim to intercept them all.


If nobody objects, I will go on with this design.,

gogonzo:
> Now, there are more shinyvalidate::InputValidator objects in this particular module and currently I am only capturing one, but I would aim to intercept them all.

Do you mean by this that you will create a hierarchy of validators using `InputValidator$add_validator`? Could you show an example how would it be organized?

nikolas-burkoff:
> Please remove all validate statements from observeEvent or observer if there are any (I saw couple somewhere)

I guess you could use a `req()` if you want the observerEvent to stop

chlebowa:
@gogonzo In the examples all of the messages were captured from one `InputValidator`:
```
iv <- shinyvalidate::InputValidator$new()
(...)
foo(iv, "Some inputs require attention")
```

I thought extending the function to be able to accept multiple validators
```
foo(iv1, iv2, "Some inputs require attention")
```
This way there would be no hierarchy, only that messages from `iv2` would appear below those from `iv1`.
[...]

PR #199, opened 2022-11-28
`shinyvalidate` improvements

chlebowa:
Closes [this issue](https://github.com/insightsengineering/teal.osprey/issues/185)

Following the introduction of `validate_inputs` to `teal` by [#199](https://github.com/insightsengineering/teal/pull/786),
this PR:

+ changes all possible input validations from `validate` calls to `shinyvalidate` input validators
+ passes all validators to `validate_input` funcitons

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -5,13 +5,14 @@
 * Added validation to `tm_g_spiderplot.R` that checks if there are duplicates in X and Y facet variables.
 * Improved input validation and used the `teal::validate_inputs` mechanism to send validation messages to the output panel.
 * Removed `purrr` from dependencies.
-* Removed redundant formal arguments from add added argument checks to `tm_g_patient_profile`.
+* Added argument checks to `tm_g_patient_profile`.
 
 ### Breaking changes
 
 * Replaced `chunks` with simpler `qenv` class.
 * Replaced `datasets` argument containing `FilteredData` with the new arguments `data` (`tdata` object) and `filter_panel_api` (`FilterPanelAPI`).
 * Updated `arm_var` to point to the factor column in `ANL`. It can't be a character column anymore.
+* Removed redundant formal arguments from `tm_g_patient_profile`.
 
 # teal.osprey 0.1.15
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,9 @@
 * Examples now use `scda.2022` instead of `scda.2021`.
 * Fixed crash in `tm_g_heat_bygrade` when not plotting `Conmed`. 
 * Added validation to `tm_g_spiderplot.R` that checks if there are duplicates in X and Y facet variables.
+* Improved input validation and used the `teal::validate_inputs` mechanism to send validation messages to the output panel.
+* Removed `purrr` from dependencies.
+* Removed redundant formal arguments from add added argument checks to `tm_g_patient_profile`.
 
 ### Breaking changes
 
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -326,8 +326,8 @@ srv_g_ae_oview <- function(id,
           code = as.expression(c(
             bquote(anl_labels <- formatters::var_labels(.(as.name(dataname)), fill = FALSE)),
             bquote(flags <- .(as.name(dataname)) %>%
-                     select(all_of(.(input$flag_var_anl))) %>%
-                     rename_at(vars(.(input$flag_var_anl)), function(x) paste0(x, ": ", anl_labels[x])))
+              select(all_of(.(input$flag_var_anl))) %>%
+              rename_at(vars(.(input$flag_var_anl)), function(x) paste0(x, ": ", anl_labels[x])))
           ))
         )
 
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -120,7 +120,10 @@ tm_g_ae_oview <- function(label,
     )
   )
   checkmate::assert_numeric(plot_height, len = 3, any.missing = FALSE, finite = TRUE)
-  checkmate::assert_numeric(plot_height[1], lower = plot_height[2], upper = plot_height[3], .var.name = "plot_height")
+  checkmate::assert_numeric(plot_height[1],
+    lower = plot_height[2], upper = plot_height[3],
+    .var.name = "plot_height"
+  )
   checkmate::assert_numeric(plot_width, len = 3, any.missing = FALSE, null.ok = TRUE, finite = TRUE)
   checkmate::assert_numeric(
     plot_width[1],
@@ -232,14 +235,35 @@ srv_g_ae_oview <- function(id,
   checkmate::assert_class(data, "tdata")
 
   moduleServer(id, function(input, output, session) {
-    iv <- shinyvalidate::InputValidator$new()
-    iv$add_rule("arm_var", shinyvalidate::sv_required())
-    iv$add_rule("flag_var_anl", shinyvalidate::sv_required(
-      message = "Please select at least one flag"
-    ))
-    iv$enable()
+    iv <- reactive({
+      ANL <- data[[dataname]]() # nolint
 
-    decorate_output <- srv_g_decorate(id = NULL, plt = plot_r, plot_height = plot_height, plot_width = plot_width)
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("arm_var", shinyvalidate::sv_required(
+        message = "Arm Variable is required"
+      ))
+      iv$add_rule("arm_var", ~ if (!is.factor(ANL[[.]])) {
+        "Arm Var must be a factor variable"
+      })
+      iv$add_rule("arm_var", ~ if (nlevels(ANL[[.]]) < 2L) {
+        "Selected Arm Var must have at least two levels"
+      })
+      iv$add_rule("flag_var_anl", shinyvalidate::sv_required(
+        message = "At least one Flag is required"
+      ))
+      rule_diff <- function(value, other) {
+        if (isTRUE(value == other)) "Control and Treatment must be different"
+      }
+      iv$add_rule("arm_trt", rule_diff, other = input$arm_ref)
+      iv$add_rule("arm_ref", rule_diff, other = input$arm_trt)
+      iv$enable()
+      iv
+    })
+
+    decorate_output <- srv_g_decorate(
+      id = NULL, plt = plot_r,
+      plot_height = plot_height, plot_width = plot_width
+    )
     font_size <- decorate_output$font_size
     pws <- decorate_output$pws
 
@@ -283,69 +307,53 @@ srv_g_ae_oview <- function(id,
       )
     })
 
-    output_q <- reactive({
-      ANL <- data[[dataname]]() # nolint
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor."
-      ))
-      validate(need(input$flag_var_anl, "Please select at least one flag."))
-      iv_comp <- shinyvalidate::InputValidator$new()
-      iv_comp$add_rule("arm_trt", shinyvalidate::sv_not_equal(
-        input$arm_ref,
-        message_fmt = "Must not be equal to Control"
-      ))
-      iv_comp$add_rule("arm_ref", shinyvalidate::sv_not_equal(
-        input$arm_trt,
-        message_fmt = "Must not be equal to Treatment"
-      ))
-      iv_comp$enable()
+    output_q <- shiny::debounce(
+      millis = 200,
+      r = reactive({
+        ANL <- data[[dataname]]() # nolint
 
-      validate(need(iv_comp$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      validate(need(nlevels(ANL[[input$arm_var]]) > 1, "Arm needs to have at least 2 levels"))
-      validate_has_data(ANL, min_nrow = 10)
-      if (all(c(input$arm_trt, input$arm_ref) %in% ANL[[input$arm_var]])) {
-        iv_an <- shinyvalidate::InputValidator$new()
-        iv_an$add_rule("arm_ref", shinyvalidate::sv_in_set(set = ANL[[input$arm_var]]))
-        iv_an$add_rule("arm_trt", shinyvalidate::sv_in_set(set = ANL[[input$arm_var]]))
-        iv_an$enable()
-        validate(need(iv_an$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      }
-      validate(need(all(c(input$arm_trt, input$arm_ref) %in% unique(ANL[[input$arm_var]])), "Plot loading"))
+        teal::validate_has_data(ANL, min_nrow = 10, msg = sprintf("%s has not enough data", dataname))
 
-      q1 <- teal.code::eval_code(
-        teal.code::new_qenv(tdata2env(data), code = get_code_tdata(data)),
-        code = as.expression(c(
-          bquote(anl_labels <- formatters::var_labels(.(as.name(dataname)), fill = FALSE)),
-          bquote(flags <- .(as.name(dataname)) %>%
-            select(all_of(.(input$flag_var_anl))) %>%
-            rename_at(vars(.(input$flag_var_anl)), function(x) paste0(x, ": ", anl_labels[x])))
+        teal::validate_inputs(iv())
+
+        validate(need(
+          input$arm_trt %in% ANL[[input$arm_var]] && input$arm_ref %in% ANL[[input$arm_var]],
+          "Treatment or Control not found in Arm Variable. Perhaps they have been filtered out?"
         ))
-      )
 
-      teal.code::eval_code(
-        q1,
-        code = as.expression(c(
-          bquote(
-            plot <- osprey::g_events_term_id(
-              term = flags,
-              id = .(as.name(dataname))[["USUBJID"]],
-              arm = .(as.name(dataname))[[.(input$arm_var)]],
-              arm_N = table(ADSL[[.(input$arm_var)]]),
-              ref = .(input$arm_ref),
-              trt = .(input$arm_trt),
-              diff_ci_method = .(input$diff_ci_method),
-              conf_level = .(input$conf_level),
-              axis_side = .(input$axis),
-              fontsize = .(font_size()),
-              draw = TRUE
-            )
-          ),
-          quote(plot)
-        ))
-      )
-    })
+        q1 <- teal.code::eval_code(
+          teal.code::new_qenv(tdata2env(data), code = get_code_tdata(data)),
+          code = as.expression(c(
+            bquote(anl_labels <- formatters::var_labels(.(as.name(dataname)), fill = FALSE)),
+            bquote(flags <- .(as.name(dataname)) %>%
+                     select(all_of(.(input$flag_var_anl))) %>%
+                     rename_at(vars(.(input$flag_var_anl)), function(x) paste0(x, ": ", anl_labels[x])))
+          ))
+        )
+
+        teal.code::eval_code(
+          q1,
+          code = as.expression(c(
+            bquote(
+              plot <- osprey::g_events_term_id(
+                term = flags,
+                id = .(as.name(dataname))[["USUBJID"]],
+                arm = .(as.name(dataname))[[.(input$arm_var)]],
+                arm_N = table(ADSL[[.(input$arm_var)]]),
+                ref = .(input$arm_ref),
+                trt = .(input$arm_trt),
+                diff_ci_method = .(input$diff_ci_method),
+                conf_level = .(input$conf_level),
+                axis_side = .(input$axis),
+                fontsize = .(font_size()),
+                draw = TRUE
+              )
+            ),
+            quote(plot)
+          ))
+        )
+      })
+    )
 
     plot_r <- reactive(output_q()[["plot"]])
 
--- a/R/tm_g_ae_sub.R
+++ b/R/tm_g_ae_sub.R
@@ -184,9 +184,33 @@ srv_g_ae_sub <- function(id,
   checkmate::assert_class(data, "tdata")
 
   moduleServer(id, function(input, output, session) {
-    iv <- shinyvalidate::InputValidator$new()
-    iv$add_rule("arm_var", shinyvalidate::sv_required())
-    iv$enable()
+    iv <- reactive({
+      ANL <- data[[dataname]]() # nolint
+      ADSL <- data[["ADSL"]]() # nolint
+
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("arm_var", shinyvalidate::sv_required(
+        message = "Arm Variable is required"
+      ))
+      iv$add_rule("arm_var", ~ if (!is.factor(ANL[[.]])) {
+        "Arm Var must be a factor variable, contact developer"
+      })
+      rule_diff <- function(value, other) {
+        if (isTRUE(value == other)) "Control and Treatment must be different"
+      }
+      iv$add_rule("arm_trt", rule_diff, other = input$arm_ref)
+      iv$add_rule("arm_ref", rule_diff, other = input$arm_trt)
+      iv$add_rule("groups", shinyvalidate::sv_in_set(
+        names(ANL),
+        message_fmt = sprintf("Groups must be a variable in %s", dataname)
+      ))
+      iv$add_rule("groups", shinyvalidate::sv_in_set(
+        names(ADSL),
+        message_fmt = "Groups must be a variable in ADSL"
+      ))
+      iv$enable()
+      iv
+    })
 
     decorate_output <- srv_g_decorate(
       id = NULL,
@@ -286,91 +310,69 @@ srv_g_ae_sub <- function(id,
       })
     })
 
-    output_q <- reactive({
-      ANL <- data[[dataname]]() # nolint
-      ADSL <- data[["ADSL"]]() # nolint
-      validate_has_data(ANL, min_nrow = 10)
-      iv_comp <- shinyvalidate::InputValidator$new()
-      iv_comp$add_rule("arm_trt", shinyvalidate::sv_not_equal(
-        input$arm_ref,
-        message_fmt = "Must not be equal to Control"
-      ))
-      iv_comp$add_rule("arm_ref", shinyvalidate::sv_not_equal(
-        input$arm_trt,
-        message_fmt = "Must not be equal to Treatment"
-      ))
-      iv_comp$enable()
-      validate(need(iv_comp$is_valid(), "Misspecification error: please observe red flags in the encodings."))
+    output_q <- shiny::debounce(
+      millis = 200,
+      r = reactive({
+        ANL <- data[[dataname]]() # nolint
+        ADSL <- data[["ADSL"]]() # nolint
 
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor. Contact the app developer."
-      ))
-      validate(
-        need(
-          all(c(input$arm_trt, input$arm_ref) %in% levels(ADSL[[input$arm_var]])),
-          "Updating treatment and control selections."
-        )
-      )
-      validate(
-        need(
-          all(c(input$arm_trt, input$arm_ref) %in% levels(ANL[[input$arm_var]])),
-          "The dataset does not contain subjects with AE events from both the control and treatment arms."
-        ),
-        need(
-          all(input$groups %in% names(ANL)) & all(input$groups %in% names(ADSL)),
-          "Check all selected subgroups are columns in ADAE and ADSL."
-        )
-      )
+        teal::validate_has_data(ANL, min_nrow = 10, msg = sprintf("%s has not enough data", dataname))
 
-      group_labels <- lapply(seq_along(input$groups), function(x) {
-        items <- input[[sprintf("groups__%s", x)]]
-        if (length(items) > 0) {
-          l <- lapply(seq_along(items), function(y) {
-            input[[sprintf("groups__%s__level__%s", x, y)]]
-          })
-          names(l) <- items
-          l[["Total"]] <- input[[sprintf("groups__%s__level__%s", x, "all")]]
-          l
-        }
-      })
+        teal::validate_inputs(iv())
 
-      group_labels_call <- if (length(unlist(group_labels)) == 0) {
-        quote(group_labels <- NULL)
-      } else {
-        bquote(group_labels <- setNames(.(group_labels), .(input$groups)))
-      }
-
-      q1 <- teal.code::eval_code(
-        teal.code::new_qenv(tdata2env(data), code = get_code_tdata(data)),
-        code = group_labels_call
-      )
-      q2 <- teal.code::eval_code(q1, code = "")
-      teal.code::eval_code(
-        q2,
-        code = as.expression(c(
-          bquote(
-            plot <- osprey::g_ae_sub(
-              id = .(as.name(dataname))$USUBJID,
-              arm = as.factor(.(as.name(dataname))[[.(input$arm_var)]]),
-              arm_sl = as.character(ADSL[[.(input$arm_var)]]),
-              trt = .(input$arm_trt),
-              ref = .(input$arm_ref),
-              subgroups = .(as.name(dataname))[.(input$groups)],
-              subgroups_sl = ADSL[.(input$groups)],
-              subgroups_levels = group_labels,
-              conf_level = .(input$conf_level),
-              diff_ci_method = .(input$ci),
-              fontsize = .(font_size()),
-              arm_n = .(input$arm_n),
-              draw = TRUE
-            )
-          ),
-          quote(plot)
+        validate(need(
+          input$arm_trt %in% ANL[[input$arm_var]] && input$arm_ref %in% ANL[[input$arm_var]],
+          "Treatment or Control not found in Arm Variable. Perhaps they have been filtered out?"
         ))
-      )
-    })
+
+        group_labels <- lapply(seq_along(input$groups), function(x) {
+          items <- input[[sprintf("groups__%s", x)]]
+          if (length(items) > 0) {
+            l <- lapply(seq_along(items), function(y) {
+              input[[sprintf("groups__%s__level__%s", x, y)]]
+            })
+            names(l) <- items
+            l[["Total"]] <- input[[sprintf("groups__%s__level__%s", x, "all")]]
+            l
+          }
+        })
+
+        group_labels_call <- if (length(unlist(group_labels)) == 0) {
+          quote(group_labels <- NULL)
+        } else {
+          bquote(group_labels <- setNames(.(group_labels), .(input$groups)))
+        }
+
+        q1 <- teal.code::eval_code(
+          teal.code::new_qenv(tdata2env(data), code = get_code_tdata(data)),
+          code = group_labels_call
+        )
+        q2 <- teal.code::eval_code(q1, code = "")
+        teal.code::eval_code(
+          q2,
+          code = as.expression(c(
+            bquote(
+              plot <- osprey::g_ae_sub(
+                id = .(as.name(dataname))$USUBJID,
+                arm = as.factor(.(as.name(dataname))[[.(input$arm_var)]]),
+                arm_sl = as.character(ADSL[[.(input$arm_var)]]),
+                trt = .(input$arm_trt),
+                ref = .(input$arm_ref),
+                subgroups = .(as.name(dataname))[.(input$groups)],
+                subgroups_sl = ADSL[.(input$groups)],
+                subgroups_levels = group_labels,
+                conf_level = .(input$conf_level),
+                diff_ci_method = .(input$ci),
+                fontsize = .(font_size()),
+                arm_n = .(input$arm_n),
+                draw = TRUE
+              )
+            ),
+            quote(plot)
+          ))
+        )
+      })
+    )
 
     plot_r <- reactive(output_q()[["plot"]])
 
--- a/R/tm_g_butterfly.R
+++ b/R/tm_g_butterfly.R
@@ -81,11 +81,11 @@
 #'       dataname = "ADAE",
 #'       right_var = choices_selected(
 #'         selected = "SEX",
-#'         choices = c("DOSE", "SEX", "ARM", "RACE", "flag1", "flag2", "flag3")
+#'         choices = c("SEX", "ARM", "RACE")
 #'       ),
 #'       left_var = choices_selected(
 #'         selected = "RACE",
-#'         choices = c("DOSE", "SEX", "ARM", "RACE", "flag1", "flag2", "flag3")
+#'         choices = c("SEX", "ARM", "RACE")
 #'       ),
 #'       category_var = choices_selected(selected = "AEBODSYS", choices = c("AEDECOD", "AEBODSYS")),
 #'       color_by_var = choices_selected(selected = "AETOXGR", choices = c("AETOXGR", "None")),
@@ -269,11 +269,35 @@ srv_g_butterfly <- function(id, data, filter_panel_api, reporter, dataname, labe
   checkmate::assert_class(data, "tdata")
 
   moduleServer(id, function(input, output, session) {
-    iv <- shinyvalidate::InputValidator$new()
-    iv$add_rule("category_var", shinyvalidate::sv_required())
-    iv$add_rule("right_var", shinyvalidate::sv_required())
-    iv$add_rule("left_var", shinyvalidate::sv_required())
-    iv$enable()
+    iv <- reactive({
+      ADSL <- data[["ADSL"]]() # nolint
+      ANL <- data[[dataname]]() # nolint
+
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("category_var", shinyvalidate::sv_required(
+        message = "Category Variable is required"
+      ))
+      iv$add_rule("right_var", shinyvalidate::sv_required(
+        message = "Right Dichotomization Variable is required"
+      ))
+      iv$add_rule("left_var", shinyvalidate::sv_required(
+        message = "Left Dichotomization Variable is required"
+      ))
+      iv$add_rule("right_var", ~ if (!is.factor(ANL[[.]])) {
+        "Right Dichotomization Variable must be a factor variable, contact developer"
+      })
+      iv$add_rule("left_var", ~ if (!is.factor(ANL[[.]])) {
+        "Left Dichotomization Variable must be a factor variable, contact developer"
+      })
+      iv$add_rule("right_val", shinyvalidate::sv_required(
+        message = "At least one value of Right Dichotomization Variable must be selected"
+      ))
+      iv$add_rule("left_val", shinyvalidate::sv_required(
+        message = "At least one value of Left Dichotomization Variable must be selected"
+      ))
+      iv$enable()
+      iv
+    })
 
     options <- reactiveValues(r = NULL, l = NULL)
     vars <- reactiveValues(r = NULL, l = NULL)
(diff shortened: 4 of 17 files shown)


R/tm_g_ae_oview.R
```diff
@@ -285,34 +281,32 @@ srv_g_ae_oview <- function(id,
...
-        "Selected arm variable needs to be a factor."
-      ))
-      validate(need(input$flag_var_anl, "Please select at least one flag."))
-      iv_comp <- shinyvalidate::InputValidator$new()
-      iv_comp$add_rule("arm_trt", shinyvalidate::sv_not_equal(
+
+      teal::validate_has_data(ANL, min_nrow = 10, msg = sprintf("%s has not enough data", dataname))
+
+      # set up and enable input validator(s)
+      iv <- shinyvalidate::InputValidator$new()
```
nikolas-burkoff, line 288: Does the validator need to be created inside a reactive - couldn't it be created in the module and then just call gather_fails() in the reactive?

chlebowa: In this particular module, yes. This is because some inputs are created in response to others. See the original version, there were actually two* validators, one created at the top of the module (line 235) and the other one in the reactive (line 294).

In some modules this is not a requirement, but in some it is a must. I decided to keep it in the reactive in all modules for consistency.

* I honestly didn't check for the third one at 309

BLAZEWIM:
Concerning gather_fails:

Like the idea and the example app. My comments / questions in the conversations.
It is only draft so I do not care about spelling or comments.
Are all those checkmate's needed? The number of developers that can make those mistakes is limited.

chlebowa:
All modules have been modified.

R/tm_g_patient_profile.R
```diff
@@ -509,6 +504,55 @@ srv_g_patient_profile <- function(id,
...
+        "Study Days Range mut be: first lower, then upper limit")
+      rule_dataset <- function(value) {
+        if (!any(c(input$select_ae, input$select_ex, input$select_rs, input$select_cm, input$select_lb)))
+          "Select at least one ADaM data set"
+      }
+      iv$add_rule("select_ae", rule_dataset)
+      iv$add_rule("select_ex", rule_dataset)
+      iv$add_rule("select_rs", rule_dataset)
+      iv$add_rule("select_cm", rule_dataset)
+      iv$add_rule("select_lb", rule_dataset)
```
chlebowa, line 549: These are five individual `checkboxInput`s so the message being displayed five times in the input (only once in the output because `gather_fails` removes duplicates). This is a little messy.  Perhaps the `checkboxInput`s could be changed into one `checkboxGroupInput`?

R/tm_g_ae_oview.R
```diff
@@ -285,34 +281,32 @@ srv_g_ae_oview <- function(id,
...
-        iv_an$add_rule("arm_trt", shinyvalidate::sv_in_set(set = ANL[[input$arm_var]]))
-        iv_an$enable()
-        validate(need(iv_an$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      }
-      validate(need(all(c(input$arm_trt, input$arm_ref) %in% unique(ANL[[input$arm_var]])), "Plot loading"))
+      # collate validator messages
+      gather_fails(iv)
+
+      validate(need(all(c(input$arm_trt, input$arm_ref) %in% unique(ANL[[input$arm_var]])),
+                    "Plot loading"))
```
nikolas-burkoff, line 309: Can we change this (I know you didn't add it in this PR but it's clearly wrong...)

chlebowa:
> Concerning gather_fails:
> 
> Like the idea and the example app. My comments / questions in the conversations. It is only draft so I do not care about spelling or comments. Are all those checkmate's needed? The number of developers that can make those mistakes is limited.

I add `checkmates` by habit in case something is ever called programmatically.

R/tm_g_ae_oview.R
chlebowa: Can you explain how it is wrong? I assumed it's a DYI spinner as it does that job rather well.

nikolas-burkoff: The plot isn't loading - there's a validation error!

R/tm_g_ae_oview.R
nikolas-burkoff: So this works outside of the reactive:

```
    # set up and enable input validator(s)
    iv <- shinyvalidate::InputValidator$new()
    iv$add_rule("arm_var", shinyvalidate::sv_required(
      message = "Arm Variable is required"))
    iv$add_rule("arm_var", ~ if (!is.factor(data[[dataname]]()[[req(.)]]))
      "Arm Var must be a factor variable")
    iv$add_rule("arm_var", ~ if (length(unique(data[[dataname]]()[[req(.)]])) < 2)
      "Selected Arm Var has not enough treatments to compare")
    iv$add_rule("flag_var_anl", shinyvalidate::sv_required(
      message = "At least one Flag is required"))
    iv$add_rule("arm_trt", ~ if (!is.null(input$arm_ref) && (.) == input$arm_ref)
      message_fmt = "Control and Treatment must be different")
    iv$add_rule("arm_ref", ~ if (!is.null(input$arm_trt) && (.) == input$arm_trt)
      message_fmt = "Control and Treatment must be different")
    iv$enable()
```

But it is a little more painful...

R/tm_g_ae_oview.R
chlebowa: I thought that was the point. 

Fair enough, fixed. 

By the way, this is one of several redundant validations. One input selects a variable and that causes another input to present this variable's available values as choices. The value selected in the second input cannot but be a element of the variable selected in the first one. Why validate it at all?

R/tm_g_ae_oview.R
chlebowa: So, three things.

1. It seems the source of the problem is the `shinyvalidate` helper functions and all we have to do is avoid them and we will be fine. Good catch!

2. This 
```
iv$add_rule("arm_trt", ~ if (!is.null(input$arm_ref) && (.) == input$arm_ref)
```
can be replaced with this
```
iv$add_rule("arm_trt", ~ if (isTRUE(. == input$arm_ref))
```
`isTRUE` and `isFALSE` saves us the trouble of testing `!is.null(.)`.

3 Unfortunately, this fails if placed outside ot the reactive:
```
iv$add_rule("arm_trt", shinyvalidate::sv_in_set(
  set = unique(data[[dataname]]()[[input$arm_var]]),
  message_fmt = "Treatment not found in Arm Variable"))
```
and so does this:
```
iv$add_rule("arm_trt", ~ if (isFALSE(. %in% unique(data[[dataname]]()[[input$arm_var]])))
  "Treatment not found in Arm Variable")
```

chlebowa: Also, when the validator is outside of the reactive, red messages appear on the widgets while the UI is being rendered, before the firs cycle is complete. Would we be fine with that? This could perhaps be prevented by adding a condition to the validator, but I'm not sure. And adding "all inputs are not NULL" is a) again, bulky and unseemly, b) may derail optional inputs.

R/gather_fails.R
```diff
@@ -0,0 +1,169 @@
...
+#'          col = input[["color"]], cex = input[["size"]])
+#'   })
+#' }
+#'
+#' if (interactive()) {
+#'   shinyApp(ui, server)
+#' }
+
+
+#' @rdname gather_fials
```
donyunardi, line 97:
```suggestion
#' @rdname gather_fails
```

R/gather_fails.R
```diff
@@ -0,0 +1,169 @@
...
+  checkmate::assert_string(header, null.ok = TRUE)
+
+  fail_messages <- gather_messages(iv)
+  failings <- add_header(fail_messages, header)
+
+  shiny::validate(shiny::need(is.null(failings), failings), ...)
+}
+
+
+#' @rdname gather_fials
```
donyunardi, line 110:
```suggestion
#' @rdname gather_fails
```

R/gather_fails.R
```diff
@@ -0,0 +1,169 @@
...
+  checkmate::assert_string(header, null.ok = TRUE)
+
+  fail_messages <- unlist(lapply(vals, gather_messages))
+  failings <- add_header(fail_messages, header)
+
+  shiny::validate(shiny::need(is.null(failings), failings))
+}
+
+
+#' @rdname gather_fials
```
donyunardi, line 124:
```suggestion
#' @rdname gather_fails
```

R/tm_g_ae_oview.R
nikolas-burkoff: Interesting - I think it's not worth moving stuff around in teal.osprey to make it work outside of the reactive - but thought it worth exploring.

I would have expected the second part of part 3. to work but 🤷 

I think the extra reactive cycle (I assume due to our use of insertUI - yet again a problem @gogonzo ) is probably what causes the red errors - if that could be fied in teal at somepoint then maybe we'd explore moving the validator outside of the reactive.

For tmc where there is a much more rigid structure, it's probably worth deciding exactly where to put the validator and to do it in the same place in all the modules

donyunardi:
Sorry, maybe I shouldn't worry about the misspelling but I thought I just pointed out so we don't forget about it.

Overall, I think this collating the validation is a great idea and will improve user's experience!

One scenario that I could recall recently is when a user got frustrated as he got into another issue after fixing one issue.
In this case, the data structure was not proper for the module (i.e. levels is expected on several columns) but I feel that expectation can be set should all validation warnings appear at once.

I think you pointed out that `gather_fails` should be in `teal` which I agree.

R/gather_fails.R
chlebowa: Thanks for catching these, probably the last place I would look for typos 🙄

R/tm_g_ae_oview.R
chlebowa: I think the main reason why things break down outside the reactive is the `updateSelectInput`s, but I have yet to confirm it.

The extra cycle may just be because the second input is updated after the first one is fully generated (a default value is properly selected). This also happens when one generates inputs server side (`selectInput` in `renderUI` and another `selectInput` in another `renderUI` dependent on that). I don't think we would be able to eliminate the extra cycles completely but putting the validator in the reactive does prevent the red  flashes (in this case).

Speaking of extra cycles, in some modules (not sure about `oview`) if you put a `browser` at the beginning of the reactive it will be triggered three times before the app fully starts. Probably nothing to worry about yet, but something to keep in mind.

chlebowa: And yes, `~ if (!is.null(input$arm_var) && !((.) %in% unique(data[[dataname]]()[[input$arm_var]])))` does work. The `!is.null` prevents a subseting error that would derail `isFALSE`. Hats off, sir.

R/tm_g_ae_oview.R
chlebowa: This can be turned into a meaningful message after all. If you filter too much, the condition WILL be broken.

I will have a look at all modules.

R/tm_g_ae_oview.R
mhallal1: Having the validation independent of reactive calls is very desirable and I dont mind the red errors upon initialization but I understand the concerns here.

chlebowa:
This PR is now blocked by [this one](https://github.com/insightsengineering/teal/pull/786).

`gather_fails` functions will be added to `teal` and this PR will use `shinyvalidate` and `gather_fails` to improve user experience.

R/tm_g_ae_sub.R
```diff
@@ -289,40 +285,41 @@ srv_g_ae_sub <- function(id,
...
+        message_fmt = "Control and Treatment must be different"
+      ))
+      iv$add_rule("groups", shinyvalidate::sv_in_set(
+        names(ANL),
+        message_fmt = "Groups must be a variable in ANL"
+      ))
+      iv$add_rule("groups", shinyvalidate::sv_in_set(
+        names(ADSL),
+        message_fmt = "Groups must be a variable in ADSL"
       ))
```
nikolas-burkoff, line 313: So I don't know whether we should do these as validations as they do depend on the data but I guess it's ok

R/tm_g_spiderplot.R
```diff
@@ -229,24 +229,48 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
...
-    iv$enable()
-
     # render plot
     output_q <- reactive({
       # get datasets ---
       ADSL <- data[["ADSL"]]() # nolint
       ADTR <- data[[dataname]]() # nolint
 
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s data has zero rows", "ADSL"))
+      teal::validate_has_data(ADTR, min_nrow = 0, msg = sprintf("%s data has zero rows", dataname))
```
nikolas-burkoff, line 239: would you put these after the shinyvalidate?

R/tm_g_spiderplot.R
```diff
@@ -229,24 +229,48 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
...
+        message = "Y Axis Variable is required"
+      ))
+      iv$add_rule("marker_var", shinyvalidate::sv_required(
+        message = "Marker Symbol Variable is required"
+      ))
+      iv$add_rule("line_colorby_var", shinyvalidate::sv_required(
+        message = "Color Variable is required"
+      ))
+      fac_dupl <- function(value, other) {
+        if (length(value) * length(other) > 0L & anyDuplicated(c(value, other))) {
```
nikolas-burkoff, line 258:
```suggestion
        if (length(value) * length(other) > 0L && anyDuplicated(c(value, other))) {
```

R/tm_g_spiderplot.R
```diff
@@ -259,9 +283,9 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
       vref_line <- input$vref_line
       href_line <- input$href_line
 
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      validate(need(nrow(ADSL) > 0, "ADSL data has zero rows"))
-      validate(need(nrow(ADTR) > 0, paste(dataname, "data has zero rows")))
+      # reference lines preprocessing
+      vref_line <- as_numeric_from_comma_sep_str(input$vref_line)
```
nikolas-burkoff, line 287: Can remove

```
vref_line <- input$vref_line
href_line <- input$href_line
```

from just above here

R/tm_g_events_term_id.R
```diff
@@ -287,34 +280,39 @@ srv_g_events_term_id <- function(id,
...
+      iv$enable()
+
+      teal::validate_inputs(iv)
 
       validate(need(
-        all(c(input$arm_trt, input$arm_ref) %in% unique(ANL[[input$arm_var]])),
-        "Cannot generate plot. The dataset does not contain subjects from both the control and treatment arms."
+        input$arm_trt %in% unique(ANL[[req(input$arm_var)]]) &&
+          input$arm_ref %in% unique(ANL[[req(input$arm_var)]]),
+        "Cannot generate plot. No subjects in both Control and Treatment arms."
```
nikolas-burkoff, line 315: I'm not sure "No subjects in both Control and Treatment arms." is quite clear I would leave the message as it was

R/tm_g_events_term_id.R
```diff
@@ -333,7 +331,7 @@ srv_g_events_term_id <- function(id,
         )
       )
 
-      validate(need(nrow(q1[["ANL"]]) > 10, "ANL needs at least 10 data points"))
+      teal::validate_has_data(q1[["ANL"]], min_nrow = 10, msg = "ANL needs at least 10 data points")
```
nikolas-burkoff, line 334: End user probably don't know what ANL is so I would change this to "Analysis dataset needs"

R/tm_g_events_term_id.R
```diff
@@ -287,34 +280,39 @@ srv_g_events_term_id <- function(id,
...
     output_q <- reactive({
       ANL <- data[[dataname]]() # nolint
 
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-
-      validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor. Contact an app developer."
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("term", shinyvalidate::sv_required(
```
nikolas-burkoff, line 291: Should we add:

```
iv$add_rule("axis", shinyvalidate::sv_required("Please select an axis side")
```

R/tm_g_butterfly.R
```diff
@@ -360,6 +354,43 @@ srv_g_butterfly <- function(id, data, filter_panel_api, reporter, dataname, labe
...
+      ))
+      iv$enable()
+
+      teal::validate_inputs(iv)
+
+      validate(
+        need(
+          input$right_val %in% ADSL[[input$right_var]] &&
+            input$left_val %in% ADSL[[input$right_var]],
+          "No observations for selected dichotomization values (filtered out?)"
```
nikolas-burkoff, line 390: So this isn't right - the example app doesn't work

R/tm_g_butterfly.R
```diff
@@ -360,6 +354,43 @@ srv_g_butterfly <- function(id, data, filter_panel_api, reporter, dataname, labe
...
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("category_var", shinyvalidate::sv_required(
+        message = "Category Variable is required"
+      ))
+      iv$add_rule("right_var", shinyvalidate::sv_required(
+        message = "Right Dichotomization Variable is required"
+      ))
+      iv$add_rule("left_var", shinyvalidate::sv_required(
+        message = "Left Dichotomization Variable is required"
+      ))
```
nikolas-burkoff, line 369: So we should also enforce that there should be at most two variables selected then we can remove the hacky code needed to enforce this at the moment

image

R/tm_g_butterfly.R
```diff
@@ -360,6 +354,43 @@ srv_g_butterfly <- function(id, data, filter_panel_api, reporter, dataname, labe
...
+      iv$add_rule("right_var", ~ if (!is.factor(ANL[[req(.)]])) {
+        "Right Dichotomization Variable must be a factor variable, contact developer"
+      })
+      iv$add_rule("left_var", ~ if (!is.factor(ANL[[req(.)]])) {
+        "Left Dichotomization Variable must be a factor variable, contact developer"
+      })
```
nikolas-burkoff, line 375: we should change the example app here (and teal.gallery?) to make sure this doesn't happen in those apps -> it does at the moment - for example:

image

R/tm_g_swimlane.R
```diff
@@ -309,26 +305,59 @@ srv_g_swimlane <- function(id,
...
+      iv$add_rule("bar_var", shinyvalidate::sv_required(
+        message = "Bar Length is required"
+      ))
+      # If reference lines are requested
+      iv$add_rule("vref_line", ~ if (anyNA(as_numeric_from_comma_sep_str(.))) {
+        "Vertical Reference Line(s) are invalid"
+      })
+      iv$enable()
+
+      teal::validate_inputs(iv)
```
nikolas-burkoff, line 352: I might put this before the data specific validations?

R/tm_g_swimlane.R
```diff
@@ -309,26 +305,59 @@ srv_g_swimlane <- function(id,
...
+          "marker-related variables need to come from marker data"
+        ))
+      }
+
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("bar_var", shinyvalidate::sv_required(
+        message = "Bar Length is required"
+      ))
+      # If reference lines are requested
+      iv$add_rule("vref_line", ~ if (anyNA(as_numeric_from_comma_sep_str(.))) {
```
nikolas-burkoff, line 347: I'm seeing lots of warnings in the console here:
image

nikolas-burkoff: Probably the case in waterfall as well?

R/tm_g_patient_profile.R
```diff
@@ -509,6 +503,65 @@ srv_g_patient_profile <- function(id,
 
     # render plot
     output_q <- reactive({
+      iv <- shinyvalidate::InputValidator$new()
```
nikolas-burkoff, line 506: I don't think the validation as is works here - as depending on what arguments the app developer gives some of the ui isn't there:
image

From:

```
library(scda)
library(nestcolor)

latest_data <- synthetic_cdisc_data("latest")

ADSL <- latest_data$adsl
ADAE <- latest_data$adae %>%
  mutate(
    ASTDT = as.Date(ASTDTM),
    AENDT = as.Date(AENDTM)
  )


ADRS <- latest_data$adrs %>%
  mutate(ADT = as.Date(ADTM))


x <- init(
  data = cdisc_data(
    cdisc_dataset("ADSL", ADSL,
      code = "ADSL <- synthetic_cdisc_data(\"latest\")$adsl"
    ),
    cdisc_dataset("ADRS", ADRS,
      code = "ADRS <- synthetic_cdisc_data(\"latest\")$adrs %>%
              mutate(ADT = as.Date(ADTM))"
    ),
    cdisc_dataset("ADAE", ADAE,
      code = "ADAE <- synthetic_cdisc_data(\"latest\")$adae %>%
              mutate(ASTDT = as.Date(ASTDTM),
                     AENDT = as.Date(AENDTM))"
    ),
    check = FALSE # set FALSE here to keep run time of example short, should be set to TRUE
  ),
  modules = modules(
    tm_g_patient_profile(
      patient_id = choices_selected(
        choices = unique(ADSL$USUBJID),
        selected = unique(ADSL$USUBJID)[1]
      ),
      sl_dataname = "ADSL",
      ae_dataname = "ADAE",
      rs_dataname = "ADRS",
      lb_dataname = NA_character_,
      ex_dataname = NA_character_,
      cm_dataname = NA_character_,
      show_ex_plot = FALSE,
      show_ae_plot = TRUE,
      show_rs_plot = TRUE,
      show_cm_plot = FALSE,
      show_lb_plot = TRUE,
      sl_start_date = choices_selected(
        selected = "TRTSDTM",
      ),
      ex_var = choices_selected(
        selected = "PARCAT2",
        choices = "PARCAT2"
      ),
      ae_var = choices_selected(
        selected = "AEDECOD",
      ),
      ae_line_col_var = choices_selected(
        selected = "AESER",
        choices = c("AESER", "AEREL")
      ),
      ae_line_col_opt = c("Y" = "red", "N" = "blue"),
      rs_var = choices_selected(
        selected = "PARAMCD",
        choices = "PARAMCD"
      ),
      cm_var = choices_selected(
        selected = "CMDECOD",
        choices = c("CMDECOD", "CMCAT")
      ),
      lb_var = choices_selected(
        selected = "LBTESTCD",
        choices = c("LBTESTCD", "LBCAT")
      ),
      x_limit = "-28, 750",
    )
  )
)
if (interactive()) {
  shinyApp(x$ui, x$server)
}
```

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +350,75 @@ srv_g_heatmap_bygrade <- function(id,
...
-        all(anno_var %in% names(ADSL)),
-        paste("Please only select annotation variable(s) in", sl_dataname, sep = " ")
+      iv$add_rule("anno_var", ~ if (length(.) > 2L) {
+        "No more than two Annotation Variables are allowed"
+      })
+      iv$add_rule("anno_var", shinyvalidate::sv_in_set(
+        set = names(ADSL),
+        message_fmt = sprintf("Study Ongoing Status must be a variable in %s", sl_dataname)
+      ))
+      iv$add_rule("anno_var", ~ if (isTRUE(input$id_var %in% .)) {
```
nikolas-burkoff, line 389: I think this message needs to be improved - surely "Id variable cannot be used as Annotation Variable" ?

image

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +350,75 @@ srv_g_heatmap_bygrade <- function(id,
...
+        message = "Visit Variable is required"
+      ))
+      iv$add_rule("ongo_var", shinyvalidate::sv_required(
+        message = "Study Ongoing Status Variable is required"
+      ))
+      iv$add_rule("ongo_var", shinyvalidate::sv_in_set(
+        set = names(ADEX),
+        message_fmt = sprintf("Study Ongoing Status must be a variable in %s", ex_dataname)
+      ))
+      iv$add_rule("ongo_var", ~ if (!is.logical(ADEX[[req(.)]])) {
```
nikolas-burkoff, line 376: Having the req() there is quite confusing? As you've put sv_required above this is it needed?

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +350,75 @@ srv_g_heatmap_bygrade <- function(id,
...
+      iv_cm <- shinyvalidate::InputValidator$new()
+      iv_cm$condition(~ isTRUE(input$plot_cm))
+      iv_cm$add_rule("conmed_var", shinyvalidate::sv_required(
+        message = "Conmed Variable is required"
+      ))
+      iv_cm$add_rule("conmed_var", shinyvalidate::sv_in_set(
+        set = names(ADCM),
+        message_fmt = sprintf("Conmed Variable must be a variable in %s", cm_dataname)
       ))
+      iv_cm$add_rule("conmed_var", ~ if (!is.factor(ADCM[[req(.)]])) {
```
nikolas-burkoff, line 406: again do you need the req() here?

R/tm_g_ae_oview.R
```diff
@@ -285,34 +284,39 @@ srv_g_ae_oview <- function(id,
...
+      iv$add_rule("arm_var", shinyvalidate::sv_required(
+        message = "Arm Variable is required"
       ))
-      validate(need(input$flag_var_anl, "Please select at least one flag."))
-      iv_comp <- shinyvalidate::InputValidator$new()
-      iv_comp$add_rule("arm_trt", shinyvalidate::sv_not_equal(
+      iv$add_rule("arm_var", ~ if (!is.factor(ANL[[req(.)]])) {
+        "Arm Var must be a factor variable"
+      })
+      iv$add_rule("arm_var", ~ if (length(levels(ANL[[req(.)]])) < 2L) {
```
nikolas-burkoff, line 297: don't think we need req here as we have arm_var is required above?

R/tm_g_ae_sub.R
```diff
@@ -289,40 +285,41 @@ srv_g_ae_sub <- function(id,
...
+
+      teal::validate_inputs(iv)
 
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
       validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor. Contact the app developer."
+        input$arm_trt %in% unique(ANL[[input$arm_var]]) ||
+          input$arm_ref %in% unique(ANL[[input$arm_var]]),
+        "Treatment or Control not found in Arm Variable. Filtered out?"
```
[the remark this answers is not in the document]
nikolas-burkoff: same for g_ae_overview

R/tm_g_events_term_id.R
nikolas-burkoff: also for g_ae_overview ?

R/tm_g_waterfall.R
```diff
@@ -293,15 +293,83 @@ srv_g_waterfall <- function(id,
...
+      iv$add_rule("href_line", ~ if (anyNA(as_numeric_from_comma_sep_str(.))) {
+        "Horizontal Reference Line(s) are invalid"
+      })
+      iv$add_rule("add_label_paramcd_rs", shinyvalidate::sv_optional())
+      iv$add_rule("add_label_paramcd_rs", shinyvalidate::sv_in_set(
+        set = adrs$PARAMCD,
+        message_fmt = "ADRS Label must be an element of ADRS PARAMCD"
+      ))
+      rule_excl <- function(value, other) {
+        if (length(value) > 0L & length(other) > 0L) {
```
nikolas-burkoff, line 340:
```suggestion
        if (length(value) > 0L && length(other) > 0L) {
```

R/tm_g_waterfall.R
```diff
@@ -293,15 +293,83 @@ srv_g_waterfall <- function(id,
...
+      iv$add_rule("anno_txt_paramcd_rs", shinyvalidate::sv_optional())
+      iv$add_rule("anno_txt_paramcd_rs", shinyvalidate::sv_in_set(
+        set = adrs$PARAMCD,
+        message_fmt = "Annotation Parameters must be elements of ADRS PARAMCD"
+      ))
+      iv$add_rule("gap_point_val", shinyvalidate::sv_optional())
+      iv$add_rule("gap_point_val", ~ if (!checkmate::test_number(suppressWarnings(as.numeric(.)), lower = 1)) {
+        "Break High Bars must be a positive number"
+      })
+      iv$add_rule("gap_point_val", ~ if (!checkmate::test_integerish(suppressWarnings(as.numeric(.)), lower = 1)) {
```
nikolas-burkoff, line 355: sv_integer() checks for integer-like I think - though is there a reason for this to be an integer? If so we should change the UI which only asks for numeric? Similarly for other test_integerish below

R/tm_g_waterfall.R
```diff
@@ -293,15 +293,83 @@ srv_g_waterfall <- function(id,
...
+      }
+      iv$add_rule("add_label_paramcd_rs", rule_excl, other = input$add_label_var_sl)
+      iv$add_rule("add_label_var_sl", rule_excl, other = input$add_label_paramcd_rs)
+      iv$add_rule("anno_txt_paramcd_rs", shinyvalidate::sv_optional())
+      iv$add_rule("anno_txt_paramcd_rs", shinyvalidate::sv_in_set(
+        set = adrs$PARAMCD,
+        message_fmt = "Annotation Parameters must be elements of ADRS PARAMCD"
+      ))
+      iv$add_rule("gap_point_val", shinyvalidate::sv_optional())
+      iv$add_rule("gap_point_val", ~ if (!checkmate::test_number(suppressWarnings(as.numeric(.)), lower = 1)) {
```
nikolas-burkoff, line 352: Could we not use `sv_gte()` or `sv_gt()`

nikolas-burkoff (commented):
Looking good - a mix of minor comments and a few more important ones e.g. butterfly plot + patient profile in particular need a bit more work

R/tm_g_ae_oview.R
chlebowa: True, must have been a holdover from when the validator was being moved around.

R/tm_g_butterfly.R
chlebowa: How is this hacky? This is a legitimate option to `selectizeInput`. Plus the flag is right in the input. Do we have to pass this through a validator?

R/tm_g_butterfly.R
chlebowa: `input$left_val %in% ADSL[[input$right_var]],` this should have been `input$left_val %in% ADSL[[input$left_var]],`
and it needed an `all` as `val`s can be length > 1.

fixed

R/tm_g_events_term_id.R
chlebowa: restoerd

R/tm_g_events_term_id.R
chlebowa: Should we? If you select nothing, the labels still appear on the left. This might be a bug in the module, I don't know.

R/tm_g_heat_bygrade.R
chlebowa: `req` removed

R/tm_g_heat_bygrade.R
chlebowa: This was the original design. It save the user the trouble of looking up what the ID Variable is nag going back to deselect it.

R/tm_g_butterfly.R
nikolas-burkoff: Personally I would much rather have the "you can only select 2 options" as a validator than the code which does this https://github.com/insightsengineering/teal.osprey/blob/main/R/tm_g_butterfly.R#L282-L319

R/tm_g_heat_bygrade.R
chlebowa: removed

R/tm_g_heat_bygrade.R
nikolas-burkoff: could we at least say why it should be removed "Deselect xxx in Annotation variables as it is the ID variable" - as on seeing this my thought was "why has the app developer let me select this here if it shouldn't be allowd"

R/tm_g_spiderplot.R
chlebowa: done

R/tm_g_spiderplot.R
chlebowa: changed to `href_line <- as_numeric_from_comma_sep_str(href_line)` to maintain original flow

R/tm_g_swimlane.R
chlebowa: These are raised by `as.numeric` in `as_numeric_from_comma_sep_str`.

R/tm_g_waterfall.R
chlebowa: This is only wrong if __both__ are length > 0.

R/tm_g_waterfall.R
chlebowa: The character string must be processed before it's tested and `sv_*` take the input value as is.

R/tm_g_waterfall.R
chlebowa: You're right, the input allows fractions, so these rules are overly stringent.

R/tm_g_butterfly.R
chlebowa: That's not what that does, though, I think.
https://github.com/insightsengineering/teal.osprey/blob/main/R/tm_g_butterfly.R#L190-L198

R/tm_g_waterfall.R
nikolas-burkoff: `length(value) > 0` and `length(other) > 0` both return a single logical value? 

Then you should use "&&" not "&"? as single "&" is the vectorized version of "&&"

image

R/tm_g_butterfly.R
nikolas-burkoff: Wonder what that does do then? And can we remove it...

chlebowa: Presents choices, I think. In a very roundabout way, which I am sure is supported by some desired behavior.

R/tm_g_ae_sub.R
```diff
@@ -289,40 +285,41 @@ srv_g_ae_sub <- function(id,
...
       ))
-      iv_comp$add_rule("arm_ref", shinyvalidate::sv_not_equal(
+      iv$add_rule("arm_ref", shinyvalidate::sv_not_equal(
         input$arm_trt,
-        message_fmt = "Must not be equal to Treatment"
+        message_fmt = "Control and Treatment must be different"
+      ))
+      iv$add_rule("groups", shinyvalidate::sv_in_set(
+        names(ANL),
+        message_fmt = "Groups must be a variable in ANL"
```
gogonzo, line 308: There is no such dataset as ANL for the end user. `ANL <- data[[dataname]]()`

```suggestion
        message_fmt = sprintf("Groups must be a variable in %s", dataname)
```

chlebowa: "ADAE" then? See line 321 in original version.

R/tm_g_ae_oview.R
```diff
@@ -285,34 +284,39 @@ srv_g_ae_oview <- function(id,
...
     output_q <- reactive({
       ANL <- data[[dataname]]() # nolint
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-      validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor."
+
+      teal::validate_has_data(ANL, min_nrow = 10, msg = sprintf("%s has not enough data", dataname))
+
+      iv <- shinyvalidate::InputValidator$new()
```
mhallal1, line 290: What do you think about moving the initialization of `iv` and its rules in a separate reactive at the top of the module and calling `teal::validate_inputs` only in `output_q`?
 This would be similar to what is introduced in tmc https://github.com/insightsengineering/teal.modules.clinical/pull/699/files and looks cleaner in a way that the validation are separate from the output call.

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +350,75 @@ srv_g_heatmap_bygrade <- function(id,
...
+      iv_cm$add_rule("conmed_var", shinyvalidate::sv_required(
+        message = "Conmed Variable is required"
+      ))
+      iv_cm$add_rule("conmed_var", shinyvalidate::sv_in_set(
+        set = names(ADCM),
+        message_fmt = sprintf("Conmed Variable must be a variable in %s", cm_dataname)
       ))
+      iv_cm$add_rule("conmed_var", ~ if (!is.factor(ADCM[[.]])) {
+        "Study Ongoing Status must be a factor variable"
+      })
```
mhallal1, line 408: Something is wrong with `conmed_var` that the validation is not rendered in the main plot space:
Screenshot from 2022-12-12 09-41-00

mhallal1 (commented):
All modules tested and working as expected, two comments to consider.

R/tm_g_ae_oview.R
```diff
@@ -285,34 +309,16 @@ srv_g_ae_oview <- function(id,
...
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor."
-      ))
-      validate(need(input$flag_var_anl, "Please select at least one flag."))
-      iv_comp <- shinyvalidate::InputValidator$new()
-      iv_comp$add_rule("arm_trt", shinyvalidate::sv_not_equal(
-        input$arm_ref,
-        message_fmt = "Must not be equal to Control"
+        input$arm_trt %in% unique(ANL[[input$arm_var]]) ||
+          input$arm_ref %in% unique(ANL[[input$arm_var]]),
```
gogonzo, line 319: `unique` here is an extra call which just slows down the execution. It's not a big deal, but I can assure you that %in% is smart enough to stop execution on the first occurrence in the vector.

```r
x <- sample(letters, size = 10000, replace = TRUE)


microbenchmark::microbenchmark(
  "a" %in% unique(x),
  "a" %in% x
)
```

```suggestion
        input$arm_trt %in% ANL[[input$arm_var]] ||
          input$arm_ref %in% ANL[[input$arm_var]],
```

R/tm_g_ae_oview.R
```diff
@@ -232,14 +235,35 @@ srv_g_ae_oview <- function(id,
...
+      ANL <- data[[dataname]]() # nolint
+
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("arm_var", shinyvalidate::sv_required(
+        message = "Arm Variable is required"
+      ))
+      iv$add_rule("arm_var", ~ if (!is.factor(ANL[[.]])) {
+        "Arm Var must be a factor variable"
+      })
+      iv$add_rule("arm_var", ~ if (length(levels(ANL[[.]])) < 2L) {
```
gogonzo, line 248:
```suggestion
      iv$add_rule("arm_var", ~ if (nlevels(ANL[[.]]) < 2L) {
```

R/tm_g_events_term_id.R
```diff
@@ -287,33 +302,18 @@ srv_g_events_term_id <- function(id,
...
-        input$arm_trt,
-        message_fmt = "Must not be equal to Treatment"
-      ))
-      iv_comp$enable()
-      validate(need(iv_comp$is_valid(), "Misspecification error: please observe red flags in the encodings."))
-
-      validate(need(
-        all(c(input$arm_trt, input$arm_ref) %in% unique(ANL[[input$arm_var]])),
+        input$arm_trt %in% unique(ANL[[req(input$arm_var)]]) &&
+          input$arm_ref %in% unique(ANL[[req(input$arm_var)]]),
```
gogonzo, line 316:
```suggestion
        input$arm_trt %in% ANL[[req(input$arm_var)]] &&
          input$arm_ref %in% ANL[[req(input$arm_var)]],
```

R/tm_g_ae_oview.R
gogonzo: I think that there should be `&&` instead `||`

R/tm_g_ae_sub.R
```diff
@@ -289,40 +313,16 @@ srv_g_ae_sub <- function(id,
...
-      validate(need(iv$is_valid(), "Misspecification error: please observe red flags in the encodings."))
+      teal::validate_has_data(ANL, min_nrow = 10, msg = sprintf("%s has not enough data", dataname))
+
+      teal::validate_inputs(iv())
+
       validate(need(
-        is.factor(ANL[[input$arm_var]]),
-        "Selected arm variable needs to be a factor. Contact the app developer."
+        input$arm_trt %in% unique(ANL[[input$arm_var]]) ||
+          input$arm_ref %in% unique(ANL[[input$arm_var]]),
```
gogonzo, line 323: we `need` both to be present in a data
```suggestion
        input$arm_trt %in% ANL[[input$arm_var]] &&
          input$arm_ref %in% ANL[[input$arm_var]],
```

R/tm_g_heat_bygrade.R
```diff
@@ -315,14 +315,83 @@ srv_g_heatmap_bygrade <- function(id,
...
+        message_fmt = sprintf("Conmed Variable must be a variable in %s", cm_dataname)
+      ))
+      iv_cm$add_rule("conmed_var", ~ if (!is.factor(ADCM[[.]])) {
+        "Study Ongoing Status must be a factor variable"
+      })
+      iv_cm$add_rule("conmed_level", ~ if (length(.) > 3L) {
+        "No more than three Conmed Levels are allowed"
+      })
+      iv_cm$enable()
+      iv_cm
```
gogonzo, line 386: Do you agree?
```suggestion
      iv_cm <- shinyvalidate::InputValidator$new()
      if (isTRUE(input$plot_cm)) {
        ADSL <- data[[sl_dataname]]() # nolint
        ADEX <- data[[ex_dataname]]() # nolint
        ADAE <- data[[ae_dataname]]() # nolint
        ADCM <- data[[cm_dataname]]() # nolint
        iv_cm$add_rule("conmed_var", shinyvalidate::sv_required(
          message = "Conmed Variable is required"
        ))
        iv_cm$add_rule("conmed_var", shinyvalidate::sv_in_set(
          set = names(ADCM),
          message_fmt = sprintf("Conmed Variable must be a variable in %s", cm_dataname)
        ))
        iv_cm$add_rule("conmed_var", ~ if (!is.factor(ADCM[[.]])) {
          "Study Ongoing Status must be a factor variable"
        })
        iv_cm$add_rule("conmed_level", ~ if (length(.) > 3L) {
          "No more than three Conmed Levels are allowed"
        })
        iv_cm$enable()
      }
      iv_cm
```

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +421,21 @@ srv_g_heatmap_bygrade <- function(id,
...
-          all(input$conmed_level %in% levels(ADCM[[input$conmed_var]])),
-          "Updating Conmed Levels"
-        ))
       }
 
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s contains no data", sl_dataname))
+
+      teal::validate_inputs(iv(), iv_cm())
+
+      validate(need(input$conmed_level %in% unique(ADCM[[input$conmed_var]]), "Updating Conmed Levels"))
```
gogonzo, line 435: This should be checked only if `isTRUE(input$plot_cm)`. I suggest to include this into `iv_cm`

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +421,21 @@ srv_g_heatmap_bygrade <- function(id,
...
-        ))
-        validate(need(
-          all(input$conmed_level %in% levels(ADCM[[input$conmed_var]])),
-          "Updating Conmed Levels"
-        ))
       }
 
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s contains no data", sl_dataname))
+
+      teal::validate_inputs(iv(), iv_cm())
```
gogonzo, line 433: it's up to you whether `isTRUE(input$plot_cm)` should be here to conditionally add `iv_cm` to the validator. Alternatively, it could be in `iv_cm` in the way I suggested there

R/tm_g_heat_bygrade.R
chlebowa: I don't, actually. What's wrong with `iv$condition()`?

R/tm_g_heat_bygrade.R
chlebowa: I don't understand. `iv_cm` is not "added to the validator", it is a separate object.

gogonzo: Please ignore this suggestion, I was wrong by thinking that whole `iv_cm()` has a rule on `isTRUE(input$plot_cm)`. It's fine now

chlebowa: It does, actually, that is why there is a separate validator object.
I think we are in agreement now.

R/tm_g_patient_profile.R
```diff
@@ -426,89 +426,112 @@ srv_g_patient_profile <- function(id,
...
 
   moduleServer(id, function(input, output, session) {
-    iv <- shinyvalidate::InputValidator$new()
-    iv$add_rule("sl_start_date", shinyvalidate::sv_required())
-    iv$add_rule("lb_var_show", shinyvalidate::sv_required())
-    iv$add_rule("ae_var", shinyvalidate::sv_required())
-    iv$enable()
-
     # only show the check box when domain data is available
-    observeEvent(ae_dataname, {
```
gogonzo, line 436: funny we had `observeEvent` on the static argument xD

R/tm_g_patient_profile.R
```diff
@@ -426,89 +426,112 @@ srv_g_patient_profile <- function(id,
...
-      if (!is.na(ae_dataname)) {
-        output$select_ae_output <- renderUI({
-          checkboxInput(
-            session$ns("select_ae"),
-            "ADAE",
-            value = !is.na(ae_dataname)
-          )
-        })
-      }
+    checkboxes <- c(ex_dataname, ae_dataname, rs_dataname, lb_dataname, cm_dataname)
```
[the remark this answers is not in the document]
chlebowa: 1. `checkboxes[!is.na(checkboxes)]` - this is a carry-over from the module design. Should I make such deep changes in this issue?
2. `show_xx_plot` are not redundant, the app developer can block a plot from display. I don't know  why they would, but they can.
3. `checkboxGroupInput` - thank you :)

R/tm_g_patient_profile.R
```diff
@@ -426,89 +426,112 @@ srv_g_patient_profile <- function(id,
...
-        })
+        updateSelectInput(
+          session,
+          "lb_var_show",
+          selected = choices_selected,
+          choices = choices
+        )
       }
-    })
+    )
```
gogonzo, line 466: observe event on checkbox[lb_dataname] is unnecessary. We should set this observer only if lb_dataname is not empty

```suggestion
        if (!is.null(lb_dataname)) {
      observeEvent(
        input$lb_var,
        {
          req(input$lb_var)
          ADLB <- data[[lb_dataname]]() # nolint
          choices <- unique(ADLB[[input$lb_var]])
          choices_selected <- if (length(choices) > 5) choices[1:5] else choices

          updateSelectInput(
            session,
            "lb_var_show",
            selected = choices_selected,
            choices = choices
          )
        }
      )
    }
```

R/tm_g_patient_profile.R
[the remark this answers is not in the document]
gogonzo: 1. Why not? Alternatively, you can set all `xx_dataname = NA` as default, because right now if anyone skips `xx_dataname` specification the module will throw an error because `c(ex_dataname, ae_dataname, ...)` will fail. But this wouldn't protect anyone from misspecifying `ex_dataname = "dawid"` as it will throw non-specific error message.
 
2. Check Ctrl+f `show_ex_plot` and see where it's used. Users can't block anything with this argument as it's not consumed at all.

R/tm_g_patient_profile.R
```diff
@@ -562,94 +578,32 @@ srv_g_patient_profile <- function(id,
...
-        if (input$select_ex == FALSE | is.na(ex_dataname)) {
-          ADEX <- NULL # nolint
-        } else {
-          ADEX <- data[[ex_dataname]]() # nolint
-          validate_has_variable(ADEX, adex_vars)
-        }
-      } else {
-        ADEX <- NULL # nolint
+      ADEX <- NULL # nolint
+      if ((ex_dataname %in% input$select_ADaM) && !is.na(ex_dataname)) {
```
gogonzo, line 582: if `ex_dataname` will be empty then `select_plot()[ex_dataname]` will never be true 
```suggestion
      if (isTRUE(select_plot()[ex_dataname])) {
```

R/tm_g_patient_profile.R
```diff
@@ -562,94 +578,32 @@ srv_g_patient_profile <- function(id,
...
-        if (input$select_ae == FALSE | is.na(ae_dataname)) {
-          ADAE <- NULL # nolint
-        } else {
-          ADAE <- data[[ae_dataname]]() # nolint
-          validate_has_variable(ADAE, adae_vars)
-        }
-      } else {
-        ADAE <- NULL # nolint
+      ADAE <- NULL # nolint
+      if ((ae_dataname %in% input$select_ADaM) && !is.na(ae_dataname)) {
```
gogonzo, line 587:
```suggestion
      if (isTRUE(select_plot()[ae_dataname]) {
```

R/tm_g_patient_profile.R
```diff
@@ -562,94 +578,32 @@ srv_g_patient_profile <- function(id,
...
-        if (input$select_cm == FALSE | is.na(cm_dataname)) {
-          ADCMD <- NULL # nolint
-        } else {
-          ADCM <- data[[cm_dataname]]() # nolint
-          validate_has_variable(ADCM, adcm_vars)
-        }
-      } else {
-        ADCM <- NULL # nolint
+      ADRS <- NULL # nolint
+      if ((rs_dataname %in% input$select_ADaM) && !is.na(rs_dataname)) {
```
gogonzo, line 592: Same as few lines above if you agree ⬆️

R/tm_g_spiderplot.R
```diff
@@ -227,26 +227,55 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
...
     # render plot
     output_q <- reactive({
       # get datasets ---
       ADSL <- data[["ADSL"]]() # nolint
       ADTR <- data[[dataname]]() # nolint
 
+      teal::validate_inputs(iv())
+
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s data has zero rows", "ADSL"))
+      teal::validate_has_data(ADTR, min_nrow = 0, msg = sprintf("%s data has zero rows", dataname))
```
gogonzo, line 277:
```suggestion
      teal::validate_has_data(ADSL, min_nrow = 1, msg = sprintf("%s data has zero rows", "ADSL"))
      teal::validate_has_data(ADTR, min_nrow = 1, msg = sprintf("%s data has zero rows", dataname))
```

R/tm_g_butterfly.R
```diff
@@ -360,6 +384,19 @@ srv_g_butterfly <- function(id, data, filter_panel_api, reporter, dataname, labe
       ADSL <- data[["ADSL"]]() # nolint
       ANL <- data[[dataname]]() # nolint
 
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s Data is empty", "ADSL"))
+      teal::validate_has_data(ANL, min_nrow = 0, msg = sprintf("%s Data is empty", dataname))
```
gogonzo, line 388:
```suggestion
      teal::validate_has_data(ADSL, min_nrow = 1, msg = sprintf("%s Data is empty", "ADSL"))
      teal::validate_has_data(ANL, min_nrow = 1, msg = sprintf("%s Data is empty", dataname))
```

R/tm_g_heat_bygrade.R
```diff
@@ -352,62 +421,21 @@ srv_g_heatmap_bygrade <- function(id,
...
-          is.factor(ADCM[[input$conmed_var]]),
-          "Conmed Variable should be a factor"
-        ))
-        validate(need(
-          all(input$conmed_level %in% levels(ADCM[[input$conmed_var]])),
-          "Updating Conmed Levels"
-        ))
       }
 
+      teal::validate_has_data(ADSL, min_nrow = 0, msg = sprintf("%s contains no data", sl_dataname))
```
gogonzo, line 431:
```suggestion
      teal::validate_has_data(ADSL, min_nrow = 1, msg = sprintf("%s contains no data", sl_dataname))
```

R/tm_g_swimlane.R
```diff
@@ -275,9 +275,18 @@ srv_g_swimlane <- function(id,
...
-    iv$add_rule("bar_var", shinyvalidate::sv_required())
-    iv$enable()
+    iv <- reactive({
+      iv <- shinyvalidate::InputValidator$new()
+      iv$add_rule("bar_var", shinyvalidate::sv_required(
+        message = "Bar Length is required"
+      ))
+      # If reference lines are requested
+      iv$add_rule("vref_line", ~ if (anyNA(suppressWarnings(as_numeric_from_comma_sep_str(.)))) {
+        "Vertical Reference Line(s) are invalid"
```
gogonzo, line 285: Message doesn't help user to figure out what is wrong exactly. Please rephrase to help

R/tm_g_spiderplot.R
```diff
@@ -227,26 +227,55 @@ srv_g_spider <- function(id, data, filter_panel_api, reporter, dataname, label,
...
+      ))
+      fac_dupl <- function(value, other) {
+        if (length(value) * length(other) > 0L && anyDuplicated(c(value, other))) {
+          "X- and Y-facet Variables must not overlap"
+        }
+      }
+      iv$add_rule("xfacet_var", fac_dupl, other = input$yfacet_var)
+      iv$add_rule("yfacet_var", fac_dupl, other = input$xfacet_var)
+      iv$add_rule("vref_line", ~ if (anyNA(suppressWarnings(as_numeric_from_comma_sep_str(.)))) {
+        "Vertical Reference Line(s) are invalid"
```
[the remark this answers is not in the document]
chlebowa: The requirements are specifically described in the input label, adding them to the error message is redundant and clutters the input.

R/tm_g_swimlane.R
```diff
@@ -309,26 +318,48 @@ srv_g_swimlane <- function(id,
...
+      } else {
+        anl <- data[[dataname]]()
+        teal::validate_has_data(anl, min_nrow = 3)
+        teal::validate_has_variable(anl, anl_vars)
+
+        validate(need(
+          !any(c(marker_pos_var, marker_shape_var, marker_color_var) %in% adsl_vars),
+          "marker-related variables need to come from marker data"
+        ))
+      }
```
gogonzo, line 355: Inputs can be validated individually.

```suggestion
      if (dataname == "ADSL") {
        iv$add_rule("bar_var", shinyvalidate::sv_in_set(
          names(ADSL),
          message_fmt = sprintf("Bar Length must be a variable in ADSL")
        ))
        iv$add_rule("bar_color_var", shinyvalidate::sv_in_set(
          names(ADSL),
          message_fmt = sprintf("Bar Color variable must be present in ADSL")
        ))
        iv$add_rule("sort_var", shinyvalidate::sv_in_set(
          names(ADSL),
          message_fmt = sprintf("Sort by variable must be present in ADSL")
        ))
        iv$add_rule("sort_var", shinyvalidate::sv_in_set(
          names(ADSL),
          message_fmt = sprintf("Annotation variables must be present in ADSL")
        ))       
      
        teal::validate_has_data(ADSL, min_nrow = 3)
        teal::validate_has_variable(ADSL, c("USUBJID", "STUDYID"))
      } else {
        anl <- data[[dataname]]()
        iv$add_rule("marker_pos_var", shinyvalidate::sv_in_set(
          names(anl),
          message_fmt = sprintf("Marker Position variable must be present in %s", dataname)
        ))
        iv$add_rule("marker_shape_var", shinyvalidate::sv_in_set(
          names(anl),
          message_fmt = sprintf("Marker Shape variable must be present in %s", dataname)
        ))
        iv$add_rule("marker_color_var", shinyvalidate::sv_in_set(
          names(anl),
          message_fmt = sprintf("Marker Color variable must be present in %s", dataname)
        ))        
        teal::validate_has_variable(anl, c("USUBJID", "STUDYID"))
        validate(need(
          !any(
            c(marker_pos_var, marker_shape_var, marker_color_var) %in% 
              unique(c(input$marker_pos_var, input$marker_shape_var, input$marker_color_var))),
          "marker-related variables need to come from marker data"
        ))
      }
```

R/tm_g_waterfall.R
```diff
@@ -292,16 +292,85 @@ srv_g_waterfall <- function(id,
...
+
+      adsl_vars <- unique(
+        c(
+          "USUBJID", "STUDYID",
+          input$bar_color_var, input$sort_var, input$add_label_var_sl, input$anno_txt_var_sl, input$facet_var
+        )
+      )
+      adtr_vars <- unique(c("USUBJID", "STUDYID", "PARAMCD", input$bar_var))
+      adrs_vars <- unique(c("USUBJID", "STUDYID", "PARAMCD", "AVALC"))
+      adrs_paramcd <- unique(c(input$add_label_paramcd_rs, input$anno_txt_paramcd_rs))
```
gogonzo, line 364: `input$bar_var` and `input$anno_txt_paramcd_rs` can be checked in the `iv` reactive

```suggestion
      adtr_vars <- unique(c("USUBJID", "STUDYID", "PARAMCD"))
      adrs_vars <- unique(c("USUBJID", "STUDYID", "PARAMCD", "AVALC"))
```

```r
       iv$add_rule("bar_var", shinyvalidate::sv_in_set(
          names(adtr),
          message_fmt = sprintf("Bar Hight variable must be present in %s", dataname_tr)
        )
       iv$add_rule("add_label_paramcd_rs", shinyvalidate::sv_in_set(
          names(adrs),
          message_fmt = sprintf("Add ADRS Label to Bars variables must be present in %s", dataname_rs)
        )
        iv$add_rule("anno_txt_paramcd_rs", shinyvalidate::sv_in_set(
          names(adrs),
          message_fmt = sprintf("Annotation Parameters variables must be present in %s", dataname_rs)
        )
```

gogonzo (approved):
👍

chlebowa:
I decided to add debouncing to `all_q` reactives in modules where the output takes long to update, which was most of them.

NEWS.md
```diff
@@ -3,6 +3,9 @@
 * Examples now use `scda.2022` instead of `scda.2021`.
 * Fixed crash in `tm_g_heat_bygrade` when not plotting `Conmed`. 
 * Added validation to `tm_g_spiderplot.R` that checks if there are duplicates in X and Y facet variables.
+* Improved input validation and used the `teal::validate_inputs` mechanism to send validation messages to the output panel.
+* Removed `purrr` from dependencies.
+* Removed redundant formal arguments from add added argument checks to `tm_g_patient_profile`.
```
nikolas-burkoff, line 8: This needs rephrasing.

Also can we have a PR in teal.gallery (earlydev app) which removes these arguments otherwise that app is broken

R/tm_g_patient_profile.R
```diff
@@ -424,647 +402,538 @@ srv_g_patient_profile <- function(id,
...
-            session$ns("select_rs"),
-            "ADRS",
-            value = !is.na(rs_dataname)
-          )
+      if (isTRUE(select_plot()[ae_dataname])) {
+        iv$add_rule("ae_var", shinyvalidate::sv_required(
+          message = "Adverse Event variable is required"
+        ))
+        iv$add_rule("ae_line_var", shinyvalidate::sv_optional())
+        iv$add_rule("ae_line_var", ~ if (length(levels(ADAE[[.]])) > length(ae_line_col_opt)) {
```
nikolas-burkoff, line 449: I don't think this is right as I get:
`Error in levels: object 'ADAE' not found`

chlebowa: Works on my end 🤔

BLAZEWIM:
One minor thing I noticed that is slightly (5%) annoying: uppercases and lowercases in labels. For example, in spiderplot, we have "Variable" used everywhere, while in patient profile it is "variable". It also goes for other packages. I noticed that developers usually just go with the convention used by labels when they create warning message, and it makes sense, but it could be managed on some lazy day.

chlebowa:
> One minor thing I noticed that is slightly (5%) annoying: uppercases and lowercases in labels. For example, in spiderplot, we have "Variable" used everywhere, while in patient profile it is "variable". It also goes for other packages. I noticed that developers usually just go with the convention used by labels when they create warning message, and it makes sense, but it could be managed on some lazy day.

I was equally annoyed by this but assumed it was due to independent design decisions by different module developers.

BLAZEWIM:
So the only problem is that the check() output is not full of `no visible binding for global variable` notes, we could use `tidyselect:: all_of`, but that will add a dependency...
if we are cool with ignoring notes, it is ok.

chlebowa:
> So the only problem is that the check() output is not full of `no visible binding for global variable` notes, we could use `tidyselect:: all_of`, but that will add a dependency... if we are cool with ignoring notes, it is ok.

We decided to ignore them.

BLAZEWIM (approved):
ok

R/gather_fails.R
```diff
@@ -0,0 +1,169 @@
...
+#' \code{gather_fails} accepts one `InputValidator`
+#' and can add a header to its validation messages.
+#' \code{gather_fails_com} accepts an arbitrary number of `InputValidator`s
+#' and prints all messages together under one header.
+#' \code{gather_fails_grp} accepts a \strong{list} of `InputValidator`s
+#' and prints messages in groups. If elements of \code{validators} are named,
+#' the names are used as headers for their respective message groups.
+#'
+#'
+#' @name gather_fails
```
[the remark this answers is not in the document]
nikolas-burkoff: hmm all of these gather_* functions should not be in the PR as they've been moved to teal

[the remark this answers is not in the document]
BLAZEWIM: Yes, those are from when it started, when I approved they got digged up.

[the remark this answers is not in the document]
chlebowa: But they aren't...

R/gather_fails.R
```diff
@@ -0,0 +1,169 @@
...
+#' This function is a hybrid solution. Given an `InputValidator` object,
+#' it extracts messages from inputs that fail validation and places them all in one
+#' validation message that is passed to a `validate`/`need` call.
+#' This way the input validator messages are repeated in the output.
+#'
+#' \code{gather_fails} accepts one `InputValidator`
+#' and can add a header to its validation messages.
+#' \code{gather_fails_com} accepts an arbitrary number of `InputValidator`s
+#' and prints all messages together under one header.
+#' \code{gather_fails_grp} accepts a \strong{list} of `InputValidator`s
```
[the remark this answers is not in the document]
chlebowa: These functions have been in `teal` for weeks now.

PR #210, opened 2023-06-27
#209 #211 substitute teal.widgets::choices_selected with teal.transform::choices_selected and fix NOTES for R CMD CHECKS

### LINKED ISSUE #209: Error: object 'choices_selected' is not exported by 'namespace:teal.widgets'
### What happened?

This [R CMD Check](https://github.com/insightsengineering/teal.osprey/actions/runs/5382918644/jobs/9769005031?pr=207#step:36:93) throws an error for `teal.widgets::choices_selected` not being an exported object from `teal.widgets`

This is the only appearance of this function in the code https://github.com/insightsengineering/teal.osprey/blob/1ce7869aebea083ef126d2a0327760788fd612e3/R/tm_g_butterfly.R#L116-L118

should this be changed to `teal.transform::select_spec` ? 

`teal.transform` is not a dependency for `teal.osprey` so I would rather found a solution within `teal.widgets`, however `teal.osprey` has `teal` in dependencies which actually loads `teal.transform`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

m7pr:
Fix #209 #211


gogonzo (commented):
There are many occurrences of  `choices_selected` (without `package::`) in roxygen examples (i don't understand why R CMD check didn't throw)

m7pr:
@gogonzo becuase teal.osprey has teal in Depends and teal has teal.transform in Depends, so once you library(teal) you get teal.transform::choices_selected visible in the global NAMESPACE as teal.transform is loaded.

R/teal.osprey.R
```diff
@@ -8,3 +8,10 @@
 #' @keywords internal
 #'
 NULL
+
+# Needed to avoid R CMD note on no visible binding
+utils::globalVariables(c(
```
pawelru, line 13: no - please don't
use `.$data` instead

pawelru: reference: https://dplyr.tidyverse.org/articles/in-packages.html?q=R%20CMD%20CHECK#data-masking-and-tidy-selection-notes

the usage of `utils::globalVariables` is too powerful and has side effects of hiding meaningful errors

chlebowa: It makes more sense to me to allow this NOTE to pass.

pawelru: If we wanted to do this correctly then yes - that would be the desired way to fix it. That's why we have a separate task for it. If we go your way (which is much faster I know) then we would have to have yet another task to convert it into `.data$` - then what's the difference to the current one? :)

I hope that this is already covered in other, actively developed packages.

m7pr: I'll try to edit the dplyr code, but if it causes the headache with NOTES maybe we should get rid of dplyr at all, utiliizing the benefit of no tidyverse dependencies, hence no possible breaking changes in the future? 

teal.osprey has around ~30 dplyr calls, but teal.goshawk has +120 dplyr calls - I wonder what's happening in there.

chlebowa: :popcorn:

m7pr: Hey @pawelru I think there is a bit more besides NSE of dplyr in here. 
Can you maybe have a look at `conmed_level` and `conmed_var` variables in `srv_g_heatmap_bygrade` inside `R/tm_g_heat_bygrade.R` and propose how to fix this?

```{R}
  srv_g_heatmap_bygrade : <anonymous>: no visible binding for global
    variable 'conmed_level'
  srv_g_heatmap_bygrade : <anonymous>: no visible binding for global
    variable 'conmed_var'
```

I think it's related to the substitution, where we later use those variables in `expr` statement
image

m7pr: in private conversation pawelru recommended to change the order of expr and env which solves this particular issue 
image

m7pr: looks like it solved most of the issues

m7pr: I'll try to see if we can live without .data$ now

gogonzo (approved):
👍

PR #212, opened 2023-06-29
extend examples timing threshold to 10 seconds

m7pr:
the same as https://github.com/insightsengineering/teal.modules.general/pull/533/


m7pr:
this also requires changes on #210 so will close this one and will contiune there

Issue #204, opened 2023-05-25
Review `tosprey` documentation before release


donyunardi:
### Summary
Related with insightsengineering/nestdevs-tasks#25
In preparation to release, we need to review/assess all documentation in the package. 

### Acceptance Criteria
* Review roxygen examples on all functions
  * Test examples
  * If it's using chunk, update to `qenv`.
* Review vignettes on these topics only:
  * Review any content around `qenv`
  * Review any content around `chunks` (i.e. make sure the right language around deprecation)
  * Review any content around shinyvalidate
* redo the content of WORDLIST
  * For instruction on how to do this, visit https://github.com/insightsengineering/coredev-tasks/issues/235
* Update NEWS
* Update README (if applicable)
* Review DESCRIPTION file (i.e check dependency versions, etc)

PR #214, opened 2023-06-30
Decouple scda

gogonzo:
closes https://github.com/insightsengineering/teal.osprey/issues/213 https://github.com/insightsengineering/teal.osprey/issues/204

see also:
- https://github.com/insightsengineering/scda/pull/125
- https://github.com/insightsengineering/scda.2022/pull/123


chlebowa (approved):
I do not mind using data sets defined in `osprey` as this package `Depends` on `osprey`, however I do recall it being said that all packages should contain *their own* data sets for tests. If that is not a crucial requirement, this is good to go.

PR #229, opened 2023-10-04
198 Include user's card labels when generating the report

kartikeyakirar:
this PR fixes https://github.com/insightsengineering/teal.osprey/issues/228

this is follow-up after https://github.com/insightsengineering/teal.reporter/pull/219

### DIFF
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -370,7 +370,7 @@ srv_g_ae_oview <- function(id,
     ### REPORTER
     if (with_reporter) {
       card_fun <- function(comment, label) {
-        card <- teal.reporter::card_template(
+        card <- teal::report_card_template(
           title = "AE Overview",
           label = label,
           with_filter = with_filter,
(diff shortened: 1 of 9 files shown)


.lintr
```diff
@@ -1,5 +1,6 @@
 linters: linters_with_defaults(
   line_length_linter = line_length_linter(120),
   cyclocomp_linter = NULL,
-  object_usage_linter = NULL
+  object_usage_linter = NULL,
+  indentation_linter = NULL
```
m7pr, line 5: @kartikeyakirar in some packages you edit that, and in some you dont?

m7pr (changes requested):
please revert lintr

.lintr
kartikeyakirar: I've made edits where the checks were failing, but I believe this issue should be addressed more thoroughly. Indentation problems are affecting most packages.

m7pr: @kartikeyakirar I rather have you update the code, so that we follow new indentation rules, instead of omitting the check.
Please remove this indentation NULL in .lintr and please edit the code where we need to remove or append lines. Sorry.
There should be a lintr function that should change indentation for you, so this should be one run of one function. Thanks

m7pr: ok we had a team call. lets leave it as it is with indentation_linter = NULL

m7pr (approved):

m7pr:
Hey @kartikeyakirar looks like this staged deps cyclic error does not appear on this PR anymore. You are good to merge

PR #232, opened 2023-11-16
Replace TealData with teal_data in the docs

### LINKED ISSUE #231: Replace TealData with teal_data in the docs
Will be needed soon to enable removal of the old classes in `teal_data`.

chlebowa:
Closes #231 

Modified example apps to use the following pattern:
1. create `cdisc_data`
2. use within to evaluate code one line at a time
3. set datanames manually (enumerate datasets)
4. set join keys automatically with `default_cdisc_join_keys[datanames(data)]`
5. assign datasets in global
6. create app
7. run app

Added package prefixes to all `teal*` package functions.

### DIFF
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -16,77 +16,51 @@
 #' @export
 #'
 #' @examples
-#' library(nestcolor)
+#' data <- cdisc_data() |>
+#'   within(library(nestcolor)) |>
+#'   within({
+#'     ADSL <- rADSL
+#'     ADAE <- rADAE
+#'     add_event_flags <- function(dat) {
+#'       dat <- dat |>
+#'         mutate(
+#'           TMPFL_SER = AESER == "Y",
+#'           TMPFL_REL = AEREL == "Y",
+#'           TMPFL_GR5 = AETOXGR == "5",
+#'           AEREL1 = (AEREL == "Y" & ACTARM == "A: Drug X"),
+#'           AEREL2 = (AEREL == "Y" & ACTARM == "B: Placebo")
+#'         )
+#'       labels <- c(
+#'         "Serious AE", "Related AE", "Grade 5 AE",
+#'         "AE related to A: Drug X", "AE related to B: Placebo"
+#'       )
+#'       cols <- c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2")
+#'       for (i in seq_along(labels)) {
+#'         attr(dat[[cols[i]]], "label") <- labels[i]
+#'       }
+#'       dat
+#'     }
+#'     ADAE <- add_event_flags(ADAE)
+#'   })
 #'
-#' ADSL <- osprey::rADSL
-#' ADAE <- osprey::rADAE
+#' datanames(data) <- c("ADSL", "ADAE")
+#' join_keys(data) <- default_cdisc_join_keys[datanames(data)]
 #'
-#' # Add additional dummy causality flags.
-#' add_event_flags <- function(dat) {
-#'   dat <- dat %>%
-#'     dplyr::mutate(
-#'       TMPFL_SER = AESER == "Y",
-#'       TMPFL_REL = AEREL == "Y",
-#'       TMPFL_GR5 = AETOXGR == "5",
-#'       AEREL1 = (AEREL == "Y" & ACTARM == "A: Drug X"),
-#'       AEREL2 = (AEREL == "Y" & ACTARM == "B: Placebo")
-#'     )
-#'   labels <- c(
-#'     "Serious AE", "Related AE", "Grade 5 AE",
-#'     "AE related to A: Drug X", "AE related to B: Placebo"
-#'   )
-#'   cols <- c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2")
-#'   for (i in seq_along(labels)) {
-#'     attr(dat[[cols[i]]], "label") <- labels[i]
-#'   }
-#'   dat
-#' }
-#' ADAE <- ADAE %>% add_event_flags()
+#' ADAE <- data[["ADAE"]]
 #'
 #' app <- init(
-#'   data = cdisc_data(
-#'     cdisc_dataset("ADSL", ADSL, code = "ADSL <- osprey::rADSL"),
-#'     cdisc_dataset("ADAE", ADAE,
-#'       code =
-#'         "ADAE <- osprey::rADAE
-#'            add_event_flags <- function(dat) {
-#'              dat <- dat %>%
-#'                dplyr::mutate(
-#'                  TMPFL_SER = AESER == 'Y',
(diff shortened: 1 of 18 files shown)


R/tm_g_ae_oview.R
```diff
@@ -16,67 +16,40 @@
 #' @export
 #'
 #' @examples
-#' library(nestcolor)
+#' data <- teal.data::cdisc_data() |>
```
gogonzo, line 19: Don't need to prefix `cdisc_data()`
```suggestion
#' data <- cdisc_data() |>
```

R/tm_g_butterfly.R
```diff
@@ -40,41 +40,33 @@
...
+#'   within(ADSL <- osprey::rADSL) |>
+#'   within(ADAE <- osprey::rADAE) |>
+#'   within(ADSL <- mutate(ADSL, DOSE = paste(sample(1:3, n(), replace = TRUE), "UG"))) |>
+#'   within(ADAE <- mutate(
+#'     ADAE,
+#'     flag1 = ifelse(AETOXGR == 1, 1, 0),
+#'     flag2 = ifelse(AETOXGR == 2, 1, 0),
+#'     flag3 = ifelse(AETOXGR == 3, 1, 0),
+#'     flag1_filt = rep("Y", n())
+#'   ))
```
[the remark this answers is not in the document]
chlebowa: I thought it's more natural to build data one expression at a time and it would be easier to understand for someone who sees `within` for the first time.

R/tm_g_patient_profile.R
```diff
@@ -62,67 +62,30 @@
...
-#'   mutate(
-#'     ADT = as.Date(ADTM),
-#'     LBSTRESN = as.numeric(LBSTRESC)
-#'   )
+#' ADSL <- data[["ADSL"]]
+#' ADAE <- data[["ADAE"]]
+#' ADCM <- data[["ADCM"]]
+#' ADRS <- data[["ADRS"]]
+#' ADEX <- data[["ADEX"]]
+#' ADLB <- data[["ADLB"]]
```
gogonzo, line 84: `ADSL` is the only binding used in code below
```suggestion
#' ADSL <- data[["ADSL"]]
```

R/tm_g_spiderplot.R
```diff
@@ -25,24 +25,22 @@
...
+#'   within(ADSL <- osprey::rADSL) |>
+#'   within(ADTR <- osprey::rADTR)
+#'
+#' teal.data::datanames(data) <- c("ADSL", "ADTR")
+#' teal.data::join_keys(data) <- teal.data::default_cdisc_join_keys[teal.data::datanames(data)]
 #'
-#' ADSL <- osprey::rADSL
-#' ADTR <- osprey::rADTR
+#' ADSL <- data[["ADSL"]]
+#' ADTR <- data[["ADTR"]]
```
gogonzo, line 39: Nothing used in the code below
```suggestion
```

R/tm_g_swimlane.R
```diff
@@ -35,37 +35,29 @@
...
+#' teal.data::datanames(data) <- c("ADSL", "ADRS")
+#' teal.data::join_keys(data) <- teal.data::default_cdisc_join_keys[teal.data::datanames(data)]
 #'
-#' ADRS <- ADRS %>%
-#'   dplyr::filter(PARAMCD == "LSTASDI" & DCSREAS == "Death") %>%
-#'   mutate(AVALC = DCSREAS, ADY = EOSDY) %>%
-#'   base::rbind(ADRS %>% dplyr::filter(PARAMCD == "OVRINV" & AVALC != "NE")) %>%
-#'   arrange(USUBJID)
+#' ADSL <- data[["ADSL"]]
+#' ADRS <- data[["ADRS"]]
```
gogonzo, line 56:
```suggestion
```

R/tm_g_ae_oview.R
chlebowa: So no `teal::` and `teal.transofrm` either?

R/tm_g_waterfall.R
```diff
@@ -47,27 +47,23 @@
...
+#'   within(ADTR <- osprey::rADTR) |>
+#'   within(ADSL$SEX <- factor(ADSL$SEX, levels = unique(ADSL$SEX)))
 #'
-#' ADSL$SEX <- factor(ADSL$SEX, levels = unique(ADSL$SEX))
+#' teal.data::datanames(data) <- c("ADSL", "ADTR", "ADRS")
+#' teal.data::join_keys(data) <- teal.data::default_cdisc_join_keys[teal.data::datanames(data)]
+#'
+#' ADSL <- data[["ADSL"]]
+#' ADTR <- data[["ADTR"]]
+#' ADRS <- data[["ADRS"]]
```
gogonzo, line 62:
```suggestion
```

R/tm_g_butterfly.R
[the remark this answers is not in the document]
gogonzo: I have opposite opinion. Each code element mixed in each line with `within` makes `code` hard to read. Let others decide on this

R/tm_g_ae_oview.R
gogonzo: Everything (teal.slice, teal.data, teal.transform) is given to the app developer by default when loading teal, it is in most of the cases concise in the `teal.module.xyz` repos, that nothing is called with namespace prefix (unless necessary). They (app developers) tend to understand `library` calls more. It has been something we had debates many times and each time had a consensus that:
- no `::` in the examples
- only `::` in the code (except `base` and `$this` package and few exceptions like `shiny`)
- **`teal::init` is an institution** and can be prefixed ;]

chlebowa: All right. This is the first time I hear about this consensus (13 months on the project).
I thought it makes sense to prefix foreign functions in examples so that the reader knows where the function comes from.

The examples (before my meddling) had `osprey::rADSL`, etc. How does that fit into this?

gogonzo: This is a grey zone, feel free to remove `osprey::` if it bothers you, I don't have anything against.

chlebowa: All or nothing.

chlebowa:
I'm afraid checks will keep failing because I neglected to put `@78_...` in my branch name.

R/tm_g_ae_oview.R
m7pr: Just passing my opinion about datasets

> I know rADSL comes from osprey that is loaded as this packaged Depends on it. However we put osprey:: and other package names in front of datasets, so that there is no conflict in the namespace. rADSL and similar like rADLB exist in a couple other packages, so prefixing a package in here makes it clear what is the source of the data.

Agree to remove `::` in front of all other functions. Just wanted to highlight we can have multiple packages loaded with the same dataname called `rADSL`. For now they are the same. I dont think they ever will be different. Just wanted to say that if user hits `?rADSL` and see a couple of packages, the prefix in the example would point him to which dataset we wanted to use. But I agree that this is a grey zone and I don't argue much to stick to my view.

PR #233, opened 2023-11-16
78 simplify joinkeys@main

### LINKED ISSUE #231: Replace TealData with teal_data in the docs

chlebowa:
Closes #231 

Modified example apps to use the following pattern:
1. create `cdisc_data`
2. use within to evaluate code
3. set datanames manually (enumerate datasets)
4. set join keys automatically with `default_cdisc_join_keys[datanames(data)]`
5. assign datasets in global where applicable
6. create app
7. run app

Removed all package prefixes in examples.

### DIFF
--- a/R/tm_g_heat_bygrade.R
+++ b/R/tm_g_heat_bygrade.R
@@ -38,8 +38,6 @@
 #'   within({
 #'     library(dplyr)
 #'     library(nestcolor)
-#'   }) |>
-#'   within({
 #'     ADSL <- rADSL %>% slice(1:30)
 #'     ADEX <- rADEX %>% filter(USUBJID %in% ADSL$USUBJID)
 #'     ADAE <- rADAE %>% filter(USUBJID %in% ADSL$USUBJID)
--- a/R/tm_g_heat_bygrade.R
+++ b/R/tm_g_heat_bygrade.R
@@ -38,14 +38,16 @@
 #'   within({
 #'     library(dplyr)
 #'     library(nestcolor)
-#'     }) |>
+#'   }) |>
 #'   within({
 #'     ADSL <- rADSL %>% slice(1:30)
 #'     ADEX <- rADEX %>% filter(USUBJID %in% ADSL$USUBJID)
 #'     ADAE <- rADAE %>% filter(USUBJID %in% ADSL$USUBJID)
 #'     ADCM <- rADCM %>% filter(USUBJID %in% ADSL$USUBJID)
 #'     # This preprocess is only to force legacy standard on ADCM
-#'     ADCM <- ADCM %>% select(-starts_with("ATC")) %>% unique()
+#'     ADCM <- ADCM %>%
+#'       select(-starts_with("ATC")) %>%
+#'       unique()
 #'     # function to derive AVISIT from ADEX
 #'     add_visit <- function(data_need_visit) {
 #'       visit_dates <- ADEX %>%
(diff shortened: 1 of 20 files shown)


R/tm_g_heat_bygrade.R
```diff
@@ -34,106 +34,81 @@
...
-#'     check = TRUE
-#'   ),
+#'     # derive ongoing status variable for ADEX
+#'     ADEX <- ADEX %>%
+#'       filter(PARCAT1 == "INDIVIDUAL") %>%
+#'       mutate(ongo_status = (EOSSTT == "ONGOING"))
+#'   })
+#'
+#' datanames(data) <- c("ADSL", "ADEX", "ADAE", "ADCM")
+#' join_keys(data) <- default_cdisc_join_keys[datanames(data)]
```
m7pr, line 78: smart :)

R/tm_g_patient_profile.R
```diff
@@ -62,70 +62,32 @@
 #' @export
 #'
 #' @examples
-#' library(nestcolor)
+#' data <- cdisc_data() |>
+#'   within(library(nestcolor)) |>
+#'   within({
+#'     ADSL <- rADSL
```
m7pr, line 68: I know `rADSL` comes from `osprey` that is loaded as this packaged Depends on it. However we put `osprey::` and other package names in front of datasets, so that there is no conflict in the namespace. `rADSL` and similar like `rADLB` exist in a couple other packages, so prefixing a package in here makes it clear what is the source of the data.

This is minor so no need to argue or discuss. This is also what's been made in other packages containing `rADSL` and `rADLB`

m7pr:
Hey, so far looks good. I will pull it locally, install and check, just in case

R/tm_g_patient_profile.R
chlebowa: See discussion [here](https://github.com/insightsengineering/teal.osprey/pull/232#discussion_r1395631626).

chlebowa:
Good point :+1:

m7pr:
Any NEWS update like in here?
https://github.com/insightsengineering/teal.modules.general/pull/597/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197ea

chlebowa:
I'm not sure, honestly. We usually only mention user-facing changes, meaning interface changes. This is just an adjustment to the larger refactor in `teal.data`.

chlebowa:
> Any NEWS update like in here? https://github.com/insightsengineering/teal.modules.general/pull/597/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197ea

That's not an enhancement...

chlebowa:
> @chlebowa should we bump version for teal.data in DESCRIPTION file?

What number, though? `main` doesn't have the changes yet, so the pipeline will fail and the feature branch has the same version number as `main` because the bump occurs on merge, so the pipeline will fail.

gogonzo (approved):
👍

m7pr:
> Any NEWS update like in here? https://github.com/insightsengineering/teal.modules.general/pull/597/files#diff-51920e95310ebfbc1ae31709f3b95f89afffbf4f1a6e38e8b2b406e2fb6197ea

> That's not an enhancement...

The way we specify data in the whole framework changed, so I think it's relevant to inform users about this change in all packages dependent on teal.data. This is a propagation of a feature from the dependent package that made us update examples. Does not hurt

PR #249, opened 2024-01-23
Append `testthat` prefix in tests

m7pr:
🥇

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -44,7 +44,7 @@ Suggests:
     nestcolor (>= 0.1.0),
     rmarkdown,
     teal.data (>= 0.3.0.9018),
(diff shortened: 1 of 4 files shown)


averissimo (approved):
Corrected the `lintr` issue and bumped {testthat} to a working version common with others.

Even 3.0.4 does not work on my machine, so played it safe with `3.1.5`

image

chlebowa:
> Even 3.0.4 does not work on my machine, so played it safe with `3.1.5`

Just make sure our CI can keep up :wink:

averissimo:
True, but it's going up to the same version as other packages (that need support for the `expect_no_error`)

Such as {teal}. This shouldn't cause problems

image

PR #258, opened 2024-03-04
257 linter exceptions@main

### LINKED ISSUE #257: global lint exclude instead of individual line exclusions
https://github.com/insightsengineering/teal.osprey/pull/250#discussion_r1511157304

chlebowa:
Closes #257.


R/tm_g_patient_profile.R
```diff
@@ -698,8 +699,8 @@ srv_g_patient_profile <- function(id,
                     ) %>%
                     select(USUBJID, ASTDT, AENDT, ASTDY, AENDY, !!quo(cm_var))
                   if (length(unique(ADCM$USUBJID)) > 0) {
-                    ADCM <- ADCM[which(ADCM$AENDY >= -28 | is.na(ADCM$AENDY) == TRUE # nolint
-                    & is.na(ADCM$ASTDY) == FALSE), ] # nolint
+                    ADCM <- ADCM[which(ADCM$AENDY >= -28 | is.na(ADCM$AENDY) == TRUE &
+                      is.na(ADCM$ASTDY) == FALSE), ]
```
averissimo, line 703: This is failing locally, it can take a single line as it's under 120 characters

```suggestion
                    ADCM <- ADCM[which(ADCM$AENDY >= -28 | is.na(ADCM$AENDY) == TRUE & is.na(ADCM$ASTDY) == FALSE), ]
```

.lintr
```diff
@@ -1,5 +1,9 @@
 linters: linters_with_defaults(
     line_length_linter = line_length_linter(120),
     cyclocomp_linter = NULL,
-    object_usage_linter = NULL
+    object_usage_linter = NULL,
+    object_name_linter = object_name_linter(
+      styles = c("snake_case", "symbols"),
+      regexes = c(ANL = "^ANL_?[0-9A-Za-z_]*$", ADaM = "^r?AD[A-Z]{2,3}_?[0-9A-Za-z]*$")
```
averissimo, line 7: I'm not sure we should include lowercases on `ADXX_yyyy` variable names for the `ADSL_p` exception, but on the other hand, why not.

Regardless, please use `{2,5}` a for ADaM rule as it allows for sponsored CDISC datasets

```suggestion
      regexes = c(
        ANL = "^ANL_?[0-9A-Za-z_]*$",
        ADaM = "^r?AD[A-Z]{2,5}_?[0-9A-Za-z_]*$"
        )
```

R/tm_g_swimlane.R
```diff
@@ -473,8 +473,7 @@ srv_g_swimlane <- function(id,
             }),
             marker_shape_opt = .(if (length(marker_shape_var) == 0) {
               NULL
-            } else if (length(marker_shape_var) > 0 &
-              all(unique(anl[[marker_shape_var]]) %in% names(marker_shape_opt)) == TRUE) { # nolint
+            } else if (length(marker_shape_var) > 0 & all(unique(anl[[marker_shape_var]]) %in% names(marker_shape_opt))) { # nolint: line_length.
```
averissimo, line 476: Alternative to consider

```suggestion
            } else if (
              length(marker_shape_var) > 0 &
                all(unique(anl[[marker_shape_var]]) %in% names(marker_shape_opt))
            ) {
```

R/tm_g_swimlane.R
```diff
@@ -486,8 +485,7 @@ srv_g_swimlane <- function(id,
             }),
             marker_color_opt = .(if (length(marker_color_var) == 0) {
               NULL
-            } else if (length(marker_color_var) > 0 &
-              all(unique(anl[[marker_color_var]]) %in% names(marker_color_opt)) == TRUE) { # nolint
+            } else if (length(marker_color_var) > 0 & all(unique(anl[[marker_color_var]]) %in% names(marker_color_opt))) { # nolint: line_length.
```
averissimo, line 488: Alternative to consider

```suggestion
            } else if (
              length(marker_color_var) > 0 &
                all(unique(anl[[marker_color_var]]) %in% names(marker_color_opt))
            ) {
```

averissimo (commented):
I get this error, when I run the linter locally.

image

I think it can be resolved as a one liner _(it is under 120 characters)_

If this is only on my laptop then ignore, I left optional comments below

R/tm_g_patient_profile.R
chlebowa: Styler did it :angry:

R/tm_g_swimlane.R
chlebowa: I would rather not grapple with styler changing the indention level.
Btw, this should be `&&`, not `&`.

R/tm_g_patient_profile.R
chlebowa: It was passing on my machine but that was before styler. I will double check.

chlebowa: Solved by one-lining.

averissimo (approved):
LGTM! :rocket:

PR #260, opened 2024-03-07
52 standardise optional notation

kartikeyakirar:
part of https://github.com/insightsengineering/nestdevs-tasks/issues/52

- standardise optional notation 
- updated roxygen tags

### DIFF
--- a/R/tm_g_heat_bygrade.R
+++ b/R/tm_g_heat_bygrade.R
@@ -7,27 +7,27 @@
 #'
 #' @inheritParams teal.widgets::standard_layout
 #' @inheritParams argument_convention
-#' @param sl_dataname (\code{character}) subject level dataset name,
-#' needs to be available in the list passed to the \code{data}
-#' argument of \code{\link[teal]{init}}
-#' @param ex_dataname (\code{character}) exposures dataset name,
-#' needs to be available in the list passed to the \code{data}
-#' argument of \code{\link[teal]{init}} \cr
-#' @param ae_dataname (\code{character}) adverse events dataset name,
-#' needs to be available in the list passed to the \code{data}
-#' argument of \code{\link[teal]{init}} \cr
-#' @param cm_dataname (\code{character}) concomitant medications dataset name,
-#' needs to be available in the list passed to the \code{data}
-#' argument of \code{\link[teal]{init}} \cr
-#' specify to \code{NA} if no concomitant medications data is available
-#' @param id_var (\code{choices_seleced}) unique subject ID variable
-#' @param visit_var (\code{choices_seleced}) analysis visit variable
-#' @param ongo_var (\code{choices_seleced}) study ongoing status variable,
-#' This variable is a derived logical variable. Usually it can be derived from \code{EOSSTT}.
-#' @param anno_var (\code{choices_seleced}) annotation variable
-#' @param heat_var (\code{choices_seleced}) heatmap variable
-#' @param conmed_var (\code{choices_seleced}) concomitant medications variable,
-#' specify to \code{NA} if no concomitant medications data is available
+#' @param sl_dataname (`character`) subject level dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()]
+#' @param ex_dataname (`character`) exposures dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()] \cr
+#' @param ae_dataname (`character`) adverse events dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()] \cr
+#' @param cm_dataname (`character`) concomitant medications dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()] \cr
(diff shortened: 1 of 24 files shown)


R/utils.R
```diff
@@ -1,13 +1,13 @@
 #' Shared Parameters
 #'
 #' @description Contains arguments that are shared between multiple functions
-#'   in the package to avoid repetition using \code{inheritParams}.
+#'   in the package to avoid repetition using `inheritParams`.
 #'
-#' @param plot_height optional, (\code{numeric}) a vector of length three with \code{c(value, min, max)}. Specifies
+#' @param plot_height (`numeric`) optional, a vector of length three with `c(value, min, max)`. Specifies
```
m7pr, line 6:
```suggestion
#' @param plot_height (`numeric`) optional vector of length three with `c(value, min, max)`. Specifies
```

R/utils.R
```diff
@@ -1,13 +1,13 @@
...
 #'   the height of the main plot.
-#' @param plot_width optional, (\code{numeric}) a vector of length three with \code{c(value, min, max)}. Specifies
+#' @param plot_width (`numeric`) optional, a vector of length three with `c(value, min, max)`. Specifies
```
m7pr, line 8:
```suggestion
#' @param plot_width (`numeric`) optional vector of length three with `c(value, min, max)`. Specifies
```

R/tm_g_heat_bygrade.R
```diff
@@ -7,27 +7,27 @@
...
+#' @param ae_dataname (`character`) adverse events dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()] \cr
+#' @param cm_dataname (`character`) concomitant medications dataset name,
+#' needs to be available in the list passed to the `data`
+#' argument of [teal::init()] \cr
+#' specify to `NA` if no concomitant medications data is available
+#' @param id_var (`choices_seleced`) unique subject ID variable
+#' @param visit_var (`choices_seleced`) analysis visit variable
+#' @param ongo_var (`choices_seleced`) study ongoing status variable,
```
m7pr, line 25:
```suggestion
#' @param ongo_var (`choices_seleced`) study ongoing status variable.
```

R/utils.R
kartikeyakirar: Hi @m7pr,as per previous conversation where you suggested adding a comma after the option. We agreed on this suggestion, So why was the comma removed here?.https://github.com/insightsengineering/nestdevs-tasks/issues/52#issuecomment-1968800721

m7pr: We only agreed to add comma after `optional` word if the sentence does not have sense with optional at the beginning.

So: 

- `optional vector of length three` has a complete sense, so no need for a comma
- `optional specifies font size` is improper English statement - this should either be `optionally specifies font size`, or `optional, specifies font size` - hence a recommendation for commas but only in a situation where without comma it does not create proper English statement.

Does it make sense?

R/tm_g_heat_bygrade.R
m7pr: @kartikeyakirar no need to create a separate commit to approve this https://github.com/insightsengineering/teal.osprey/pull/260/commits/589902cbb6dd248dbbf63883587f0b7a0c8ac727

You can always `Sign off and commit suggestion` to accept this suggestion directly from GitHub.

image


The only motivation that I see to create a commit instead of accepting a suggestion is to have a documentation be rendered. But you did not render the documentation in your commit.

R/utils.R
kartikeyakirar: okay 👍

m7pr: Since `optional` is an adjective you do not need to put comma after `optional` if it comes before the `noun`. However if it comes before the `verb` you need to have a comma. Adjectives are meant to describe nouns. For verbs you either change `optional` to `optionally` or you put comma before the verb. 

Is it more clear?

R/tm_g_heat_bygrade.R
kartikeyakirar: > You can always Sign off and commit suggestion to accept this suggestion directly from GitHub.

I did same. thats why on this commit, you were shown as a co-author . on separate commit Github doesnt' show.
image

R/utils.R
kartikeyakirar: yes, my understanding on this was while `optional` serves as a notation rather than a textual addition to the sentence. the user knows it's optional, so need to be part of description.  I am okay,  it being part of a sentence in suggested cases.

kartikeyakirar:
@m7pr document inherit description from teal.widgets. so this needs to be merged and re-triggered checks after https://github.com/insightsengineering/teal.widgets/pull/231 is merged.

R/tm_g_heat_bygrade.R
m7pr: coolio. my bad

m7pr:
Yes, I noticed. Feel free to merge https://github.com/insightsengineering/teal.widgets/pull/231
then rebuild docs in here. And then merge this one.

PR #276, opened 2024-06-19
Fixing bugs

kartikeyakirar:
this is part of https://github.com/insightsengineering/coredev-tasks/issues/567
Here bugs are resolved found here

- [x] resolved delayed data cases based on https://github.com/insightsengineering/coredev-tasks/issues/558
- [x] Added get_choices function.

### DIFF
--- a/R/tm_g_patient_profile.R
+++ b/R/tm_g_patient_profile.R
@@ -368,16 +368,14 @@ srv_g_patient_profile <- function(id,
       vapply(checkboxes, function(x) x %in% input$select_ADaM, logical(1L))
     )
 
-    isolate({
-      resolved <- teal.transform::resolve_delayed(patient_id, as.list(data()@env))
+    resolved <- teal.transform::resolve_delayed(patient_id, as.list(isolate(data())@env))
 
-      updateSelectizeInput(
-        session = session,
-        inputId = "patient_id",
-        choices = resolved$choices,
-        selected = resolved$selected
-      )
-    })
+    updateSelectizeInput(
+      session = session,
+      inputId = "patient_id",
+      choices = resolved$choices,
+      selected = resolved$selected
+    )
 
     if (!is.na(lb_dataname)) {
       observeEvent(input$lb_var, ignoreNULL = TRUE, {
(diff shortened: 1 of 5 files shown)


R/tm_g_spiderplot.R
```diff
@@ -236,14 +239,27 @@ ui_g_spider <- function(id, ...) {
...
+      env <- as.list(data()@env)
+      resolved_paramcd <- teal.transform::resolve_delayed(paramcd, env)
+
+      teal.widgets::updateOptionalSelectInput(
+        session = session,
+        inputId = "paramcd",
+        choices = resolved_paramcd$choices,
+        selected = resolved_paramcd$selected
+      )
+    })
```
vedhav, line 261: Can we add explicit `isolate` to only places where this is intended for?
Because, this will make it clear what we're trying to isolate. In this case it looks like we need to stop reactivity on `data()` so it would be nice if we could just call a `isolate(data())` instead.

R/tm_g_patient_profile.R
```diff
@@ -191,6 +191,7 @@ tm_g_patient_profile <- function(label = "Patient Profile Plot",
     ui_args = args,
     server = srv_g_patient_profile,
     server_args = list(
+      patient_id = patient_id,
```
vedhav, line 194: Additionally, the module also seems to use these boolean args:
```r
show_ae_plot = TRUE,
show_rs_plot = TRUE,
show_cm_plot = FALSE,
show_lb_plot = FALSE,
```

But, I think we can just remove them. It seems like a remnant of the past.

vedhav (commented):
Added one minor comment regarding `isolate`. I can confirm that all the modules work fine in the upversioning app.

vedhav (approved):

R/tm_g_spiderplot.R
vedhav: Please make sure to isolate in other places affected in this PR too :)

R/tm_g_patient_profile.R
kartikeyakirar: yes. I have removed from upversioning app module.

R/tm_g_spiderplot.R
kartikeyakirar: Good catch, Vedha. I have updated it as per your suggestion.
Commit: 4cb9f0b

PR #295, opened 2025-01-30
Remove lookup-refs, vbump dependencies + add Remotes

m7pr:
Part of https://github.com/insightsengineering/coredev-tasks/issues/609

From now on, we will provide development dependencies in 
```
Remotes: repo/project@branch
```
format, so it's explicitly visible in the DESCRIPTION file and can be handled by `pak::install`, `renv::install` and `remotes::install`.

With development dependencies specified in CJ Pipelines configuration, this connection was hidden, and it was hard to install the package from the main branch (or any other branch) locally from user's machine.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -47,6 +47,7 @@ Suggests:
     testthat (>= 3.1.5),
     withr (>= 2.0.0)
 Remotes:
+    insightsengineering/osprey,
     insightsengineering/teal,
     insightsengineering/teal.transform,
     insightsengineering/teal.widgets
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -37,7 +37,7 @@ Imports:
     teal.data (>= 0.7.0),
     teal.logger (>= 0.3.1),
     teal.reporter (>= 0.4.0),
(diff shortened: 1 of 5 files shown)


DESCRIPTION
```diff
@@ -46,6 +46,11 @@ Suggests:
     rmarkdown (>= 2.23),
     testthat (>= 3.1.5),
     withr (>= 2.0.0)
+Remotes:
+    insightsengineering/osprey,
+    insightsengineering/teal,
+    insightsengineering/teal.transform,
+    insightsengineering/teal.widgets
```
pawelru, line 53: do we need those?
maybe wait for a realease if that's a matter of few days and merge without it?

m7pr: @pawelru I rather get those now, and see proper builds passing.
and then get this removed later. I think teal+ teal.transform still might be a week or longer

m7pr:
cc @pawelru @llrs-roche I vbumped teal.widgets to it's version released on CRAN

llrs-roche (approved):
teal and teal.transform are both on x.y.z.dev version and appear on Remotes. Osprey is als on Remotes as it is not on CRAN or any other major repository.

PR #305, opened 2025-06-19
fix failed jobs

m7pr:
Fixes https://github.com/insightsengineering/teal.osprey/issues/303
Worfklows

- 1 https://github.com/insightsengineering/teal.osprey/actions/runs/15757238218
- 2 https://github.com/insightsengineering/teal.osprey/actions/runs/15777476701


llrs-roche (approved):
Runs ok, the teal.logger version can't be installed on release which is expected and thus causes this strategy to fail.

PR #321, opened 2025-10-14
restructure sections

m7pr:
Part of:
- https://github.com/insightsengineering/teal.reporter/issues/400

Restructured sections-headers. Pleaase check out with

- https://github.com/insightsengineering/teal.reporter/pull/425
- https://github.com/insightsengineering/teal/pull/1634

### DIFF
--- a/R/tm_g_ae_oview.R
+++ b/R/tm_g_ae_oview.R
@@ -279,7 +279,6 @@ srv_g_ae_oview <- function(id,
         obj <- data()
         teal.reporter::teal_card(obj) <-
           c(
-            teal.reporter::teal_card("# AE Overview"),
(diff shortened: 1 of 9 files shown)


llrs-roche (approved):
Table of contents look ok:

Image

I couldn't find any other section titles to modify.

----

However, I noticed that it doesn't have any code to load the data for the app. The report starts with the module without any code showing how the data is loaded.

llrs-roche:
If you run the first 4 lines of code (without the pipe) on a fresh R session you'll see that it generates an error. It lacks loading a library with the rADRS being loaded. rADRS is present on osprey but the library is not loaded or available, so the code is not reproducible.

PR #329, opened 2026-03-05
`picks` version for `tm_g_events_term_id`

m7pr:
image

```r
devtools::load_all("../teal")
devtools::load_all("../teal.data")
devtools::load_all("../osprey")
devtools::load_all("../teal.picks")
devtools::load_all(".")

data <- teal_data() %>%
  within({
    ADSL <- rADSL
    ADAE <- rADAE
  })

join_keys(data) <- default_cdisc_join_keys[names(data)]

app <- init(
  data = data,
  modules = modules(
    tm_g_events_term_id(
      label = "Common AE (picks)",
      term_var = picks(
        datasets("ADAE"),
        variables(
          choices = is_categorical(min.len = 2),
          selected = "AEDECOD",
          multiple = FALSE
        )
      ),
      arm_var = picks(
        datasets("ADSL"),
        variables(
          choices = is_categorical(min.len = 2),
          selected = "ACTARMCD",
          multiple = FALSE
        )
      ),
      plot_height = c(600, 200, 2000)
    )
  )
)

shiny::shinyApp(app$ui, app$server)
```

PR #331, opened 2026-03-12
Add picks variable selection to ae_oview module

### LINKED ISSUE #330: [Feature Request]: Allow modules to use teal.picks
### Feature description

With the creation of teal.picks repository modules can be refactored so they also use teal.picks. Here is a list of modules and related PRs:

|module | assigne | PR | Status |
|--------------|-------------|--------|------|
|tm_g_events_term_id| @osenan | https://github.com/insightsengineering/teal.osprey/pull/341|Merged |
|tm_g_ae_oview| @osenan | https://github.com/insightsengineering/teal.osprey/pull/331| Merged |
|tm_g_ae_sub | @llrs-roche | https://github.com/insightsengineering/teal.osprey/pull/339 | Merged |
|tm_g_butterfly | @osenan| https://github.com/insightsengineering/teal.osprey/pull/335| Merged|
|tm_g_decorate| | Not a module to migrate|
|tm_g_heat_by_grade|@osenan |https://github.com/insightsengineering/teal.osprey/pull/337 | Merged |
|tm_g_patient_profile|@osenan  |https://github.com/insightsengineering/teal.osprey/pull/342 | Merged|
|tm_g_spiderplot| @osenan | https://github.com/insightsengineering/teal.osprey/pull/336| Merged |
|tm_g_swimlane| @osenan | https://github.com/insightsengineering/teal.osprey/pull/347 | Approved|
|tm_g_waterfall  | @averissimo  |https://github.com/insightsengineering/teal.osprey/pull/345 |Merged |

osenan:
# Pull Request

This partially closes #330 

I based my changes [in this PR](https://github.com/insightsengineering/teal.osprey/pull/329).

Please compare both modules with this example app (requires to have teal.picks installed, osprey)

```
library("teal")
library("teal.picks")
devtools::load_all()

data <- teal_data() %>%
  within({
    library(dplyr)
    ADSL <- teal.data::rADSL
    ADAE <- teal.data::rADAE
    .add_event_flags <- function(dat) {
      dat <- dat %>%
        mutate(
          TMPFL_SER = AESER == "Y",
          TMPFL_REL = AEREL == "Y",
          TMPFL_GR5 = AETOXGR == "5",
          AEREL1 = (AEREL == "Y" & ACTARM == "A: Drug X"),
          AEREL2 = (AEREL == "Y" & ACTARM == "B: Placebo")
        )
      labels <- c(
        "Serious AE", "Related AE", "Grade 5 AE",
        "AE related to A: Drug X", "AE related to B: Placebo"
      )
      cols <- c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2")
      for (i in seq_along(labels)) {
        attr(dat[[cols[i]]], "label") <- labels[i]
      }
      dat
    }
    ADAE <- .add_event_flags(ADAE)
  })

app <- init(
  data = data,
  modules = modules(
    tm_g_ae_oview(
      label = "Common AE (picks)",
      flag_var_anl = teal.picks::picks(
        teal.picks::datasets("ADAE"),
        teal.picks::variables(
           choices = c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2"),
          selected = "TMPFL_SER"
        )
      ),
      arm_var = teal.picks::picks(
        teal.picks::datasets("ADSL"),
        teal.picks::variables(
          choices = teal.picks::is_categorical(min.len = 2),
          selected = "ACTARMCD"
        )
      ),
    ),
    # Default method (old) for comparison
    tm_g_ae_oview(
      label = "Common AE (default)",
      dataname = "ADAE",
      flag_var_anl = teal.transform::choices_selected(
        selected = "AEREL1",
        choices = teal.transform::variable_choices(
          data[["ADAE"]],
          c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2")
        )
      ),
      arm_var = teal.transform::choices_selected(
        selected = "ACTARMCD",
        choices = c("ACTARM", "ACTARMCD")
      ),
    )
  )
)

shinyApp(app$ui, app$server)
```

I refactored the `tm_g_ae_oview` module to create a generic, keep the default for `teal.transform` and add a method for picks variable selection. The module works as expected and setting the variables is better with picks. The issue reported in the similar PR on the plot dimensions is fixed if we set the plot dimensions in the module. The ui of the picks ui components is different and it breaks, in my opinion, the color pattern of the teal app.

image
image

Here the teal.transform module

image
image

I added simple tests for the module as well


osenan:
I have read the CLA Document and I hereby sign the CLA

osenan:
Changed to draft as I need to fix:
- [ ] Checks 
- [ ] Plot
- [ ] Decorators

R/tm_g_ae_oview.R
```diff
@@ -73,14 +79,40 @@
...
+                              selected = 1L
+                            )
+                          ),
+                          flag_var_anl = teal.picks::picks(
+                            teal.picks::datasets(),
+                            teal.picks::variables(
+                              choices = teal.picks::is_categorical(min.len = 2),
+                              selected = 1L
+                            )
+                          ),
```
averissimo: I think this is a good candidate to use `variables()` in the arguments:

```suggestion
                          arm_var = teal.picks::variables(
                            choices = teal.picks::is_categorical(min.len = 2),
                            selected = 1L
                          ),
                          flag_var_anl = teal.picks::variables(
                            choices = teal.picks::is_categorical(min.len = 2),
                            selected = 1L
                          ),
```

You can then overwrite arguments with `picks()` in `tm_g_ae_oview.variables` (and use `dataname`):

```r
# ...
   arm_var <- teal.picks::picks(datasets(dataname), arm_var)
   flag_var_anl <- teal.picks::picks(datasets(dataname), flag_var_anl)
```

DESCRIPTION
```diff
@@ -50,9 +51,11 @@ Suggests:
     withr (>= 3.0.0)
 Remotes:
     insightsengineering/osprey
+    insightsengineering/teal.picks@redesign_exptraction@main
```
averissimo:
```suggestion
insightsengineering/osprey,
    insightsengineering/teal.picks@redesign_extraction@main
```

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+    server = srv_g_ae_oview.picks,
+    server_args = args[names(args) %in% names(formals(srv_g_ae_oview.picks))],
+    ui = ui_g_ae_oview.picks,
+    ui_args = args[names(args) %in% names(formals(ui_g_ae_oview.picks))],
+    transformators = transformators,
+    datanames = .picks_datanames(list(arm_var, flag_var_anl))
+  )
+}
+
+ui_g_ae_oview.picks <- function(
```
averissimo: No need to set default here, using dots should be fine.

```r
ui_g_ae_oview.picks <- function(id, ...) {
  a <- rlang::dots_list(...) # module args
  ns <- NS(id)
  # ...
```

But if you prefer, to keep it this way, I don't think we need to set defaults here, they should be inherited from `tm_g_ae_oview()` call

R/tm_g_ae_oview.R
```diff
@@ -73,14 +79,40 @@
...
-                          flag_var_anl,
+                          dataname = NULL,
+                          arm_var = teal.picks::picks(
+                            teal.picks::datasets(),
+                            teal.picks::variables(
+                              choices = teal.picks::is_categorical(min.len = 2),
+                              selected = 1L
+                            )
+                          ),
+                          flag_var_anl = teal.picks::picks(
```
averissimo: Should we define the defaults from arguments other than `arm_var`? it seems redundant.

Maybe keep only `arm_var` and then define the specific ones in the methods.

On the other hand, if we do this, the main function declaration doesn't have any defaults shown in the documentation. 

WDYT?

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+          "Note: %d%% CI is calculated using %s",
+          round(conf_level * 100),
+          name_ci(diff_ci_method)
+        )
+      )
+    })
+
+    observeEvent(selectors$arm_var(), {
+      arm_var_name <- selectors$arm_var()$variables$selected
+      arm_dataset <- selectors$arm_var()$datasets$selected
```
averissimo: Shouldn't this always be `dataname`?

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+        value = sprintf(
+          "Note: %d%% CI is calculated using %s",
+          round(conf_level * 100),
+          name_ci(diff_ci_method)
+        )
+      )
+    })
+
+    observeEvent(selectors$arm_var(), {
+      arm_var_name <- selectors$arm_var()$variables$selected
```
averissimo: I'm not sure here the best practice on getting the selected, I'm leaning towards using the `var` from `merged`

```suggestion
    observeEvent(merged$variables()$arm_var, {
      arm_var_name <- merged$variables()$arm_var
```

However, I don't see an issue with your approach.

WDYT?

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+      selectInput(
+        ns("arm_ref"),
+        "Control",
+        choices = NULL
+      ),
+      selectInput(
+        ns("arm_trt"),
+        "Treatment",
+        choices = NULL
+      ),
```
averissimo: Interesting case here as we have dependent input fields from `arm_var = picks(...)`.

Should these be created independently reusing `arm_var` datasets & variables

- that is: having 2 picks
- this will create problems downstream though

Or [possible feature]

Is this a case for having multiple `values()` in a picks, which generate multiple selectors.

```r
picsk(datasets("ADAE"), variables(), values(), values()) # with some way to differentiate between control and treatement
```

@gogonzo maybe this is one of the cases of `bookmark manager` we remember.

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+#'       ),
+#'       plot_height = c(600, 200, 2000)
+#'     )
+#'   )
+#' )
+#' if (interactive()) {
+#'   shinyApp(app$ui, app$server)
+#' }
+#' @export
+tm_g_ae_oview.picks <- function(label, # nolint: object_name_linter
```
averissimo:
```suggestion
tm_g_ae_oview.picks <- function(label, # nolint: object_name_linter.
```

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+    server_args = args[names(args) %in% names(formals(srv_g_ae_oview.picks))],
+    ui = ui_g_ae_oview.picks,
+    ui_args = args[names(args) %in% names(formals(ui_g_ae_oview.picks))],
+    transformators = transformators,
+    datanames = .picks_datanames(list(arm_var, flag_var_anl))
+  )
+}
+
+ui_g_ae_oview.picks <- function(
+  # nolint: object_name_linter
```
averissimo:
```suggestion
ui_g_ae_oview.picks <- function(# nolint: object_name_linter.
```

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+        ns(NULL),
+        fontsize = fontsize,
+        titles = "AE Overview",
+        footnotes = ""
+      )
+    )
+  )
+}
+
+srv_g_ae_oview.picks <- function(id, # nolint: object_name_linter
```
averissimo:
```suggestion
srv_g_ae_oview.picks <- function(id, # nolint: object_name_linter.
```

R/tm_g_ae_oview_picks.R
```diff
@@ -0,0 +1,376 @@
...
+  args <- as.list(environment())
+
+  module(
+    label = label,
+    server = srv_g_ae_oview.picks,
+    server_args = args[names(args) %in% names(formals(srv_g_ae_oview.picks))],
+    ui = ui_g_ae_oview.picks,
+    ui_args = args[names(args) %in% names(formals(ui_g_ae_oview.picks))],
+    transformators = transformators,
+    datanames = .picks_datanames(list(arm_var, flag_var_anl))
```
averissimo: Is this really needed? Or can we just use `dataname`

averissimo (commented):
Some preliminary comments.

- I believe this is a good case for only keeping 1 version of the module logic. (We can convert directly from `choices_selected` -> `variables()`)
- I see that you're using `picks()`, consider using `variables()` instead for `arm_var` and `flag_var_anl`
  - And then create `picks()` inside the `tm_g_ae_oview()` call

R/tm_g_ae_oview_picks.R
llrs-roche: This comes from the "reference" implementation at #329 but maybe we first need to review it and merge before we propagate the errors to multiple PR as I have used the same on another PR...

R/tm_g_ae_oview.R
osenan: Hi, I agree the current implementation is not coherent. I think you do not define defaults at the generic and define them at the S3 methods. No problem if they look empty at the documentation for the generic

R/tm_g_ae_oview_picks.R
osenan: This function is also at [tmg PR on picks ](https://github.com/insightsengineering/teal.modules.general/pull/942) II do not fully understand why it is necessary

llrs-roche: I see there's an issue for this now: https://github.com/insightsengineering/teal.picks/issues/17 Let's continue the discussion there as it affects multiple packages not just this one.

averissimo: Typically on `tmg` there is less control on the datasets being used, it can use the same or a merge from different ones.

In `tmc` and other modules we have a `dataname` argument and existing implementation uses that to generate the `data_extract_spec`, hence the use of `choices_selected` in arguments.

I questioned this line because we have that information from `dataname` in the arguments and we don't need to query the `picks`

R/tm_g_ae_oview.R
averissimo: Oki doki! Let's do that then, keep the defaults in the methods! _(with exception of the variable that determines the dispatch)_

R/tm_g_ae_oview_picks.R
averissimo: By this I mean `arm_dataset`

R/tm_g_ae_oview_picks.R
llrs-roche: I think this is related to this issue: https://github.com/insightsengineering/teal.picks/issues/14.  
My initial plan was to go for the second option. But it is hard to distinguish here what is what: should `picks(datasets("ADAE"), variables(), values(), values())` join all the values in a single selection or not? To disambiguate I thought about using operators: `picks(datasets("ADAE"), variables()) + values() | values()`: a picks with two options/selectors. 

Currently I think it is better if we create a function to handle those cases.

R/tm_g_ae_oview_picks.R
llrs-roche: Thanks for the explanation @averissimo. I think this makes a point against implementing such a function then.

osenan:
I have read the CLA Document and I hereby sign the CLA

R/tm_g_ae_oview.R
osenan: Changed, actually it is simpler good idea

R/tm_g_ae_oview_picks.R
osenan: Hey, due to the way the ui argument is called using `formals`, I need to keep arguments defined in the ui function. But I have changed it to not set defaults on the ui function

R/tm_g_ae_oview_picks.R
osenan: I have removed the tmg function. Many thanks! for the insight. It works correctly as well just directly using the dataname

R/tm_g_ae_oview_picks.R
osenan: It will always be the `dataname` but given that we have the funcionality in picks, I would leave it as it is so the var name and the dataname are obtained trough `picks`

R/tm_g_ae_oview_picks.R
osenan: Let's solve the issue and then update this module

osenan:
The plot and decorator I will handle in a separate issue finally

osenan:
Now the example should be tested with this app:
```

library(teal)
devtools::load_all()
library(teal.picks)

data <- teal_data() %>%
   within({
     library(dplyr)
     ADSL <- rADSL
     ADAE <- rADAE
     .add_event_flags <- function(dat) {
       dat <- dat %>%
         mutate(
           TMPFL_SER = AESER == "Y",
           TMPFL_REL = AEREL == "Y",
           TMPFL_GR5 = AETOXGR == "5",
           AEREL1 = (AEREL == "Y" & ACTARM == "A: Drug X"),
           AEREL2 = (AEREL == "Y" & ACTARM == "B: Placebo")
         )
       labels <- c(
         "Serious AE", "Related AE", "Grade 5 AE",
         "AE related to A: Drug X", "AE related to B: Placebo"
       )
       cols <- c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2")
       for (i in seq_along(labels)) {
         attr(dat[[cols[i]]], "label") <- labels[i]
       }
       dat
     }
     ADAE <- .add_event_flags(ADAE)
   })

 ADAE <- data[["ADAE"]]

 app <- init(
   data = data,
   modules = modules(
     tm_g_ae_oview(
       label = "AE Overview",
       dataname = "ADAE",
       arm_var = teal.picks::variables(
          choices = tidyselect::starts_with("ACTARM"),
          selected = "ACTARMCD"
      ),
      flag_var_anl = teal.picks::variables(
          choices = c("TMPFL_SER", "TMPFL_REL", "TMPFL_GR5", "AEREL1", "AEREL2"),
          selected = "AEREL1"
      ),
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }
```

R/tm_g_ae_oview.R
```diff
@@ -73,14 +79,28 @@
...
                           flag_var_anl,
-                          fontsize = c(5, 3, 7),
-                          plot_height = c(600L, 200L, 2000L),
-                          plot_width = NULL,
-                          transformators = list()) {
+                          fontsize,
+                          plot_height,
+                          plot_width,
+                          transformators) {
+  UseMethod("tm_g_ae_oview", flag_var_anl)
```
averissimo: Why not use `arm_var` as the variable?

R/tm_g_ae_oview.R
```diff
@@ -295,13 +315,13 @@ srv_g_ae_oview <- function(id,
           "Treatment or Control not found in Arm Variable. Perhaps they have been filtered out?"
         ))
 
-        q1 <- obj %>%
+        q1 <- obj |>
```
averissimo: This is not supported by `R >= 3.6`. To do this we need to bump the version of R in the `DESCRIPTION` to at least `4.1` (4.3 to get `%||%` should be safe)

I'm not opposed to bumping it, but if this is the only reason I would leave it as is.

R/tm_g_ae_oview.R
```diff
@@ -73,14 +79,28 @@
...
+#' @export
+tm_g_ae_oview.default <- function(label,
+                                  dataname,
+                                  arm_var,
+                                  flag_var_anl,
+                                  fontsize = c(5, 3, 7),
+                                  plot_height = c(600L, 200L, 2000L),
+                                  plot_width = NULL,
+                                  transformators = list()) {
   message("Initializing tm_g_ae_oview")
```
averissimo, line 81: Please use picks back-end directly here and don't keep any other logic here, save some validation. `ui_*` and `srv_*` can both be deleted

This works flawlessly 🥳 

```suggestion
  message("Initializing tm_g_ae_oview")
  checkmate::assert_class(arm_var, classes = "choices_selected")
  checkmate::assert_class(flag_var_anl, classes = "choices_selected")
  tm_g_ae_oview.pick(
    label,
    dataname,
    teal.picks::as.picks(arm_var),
    teal.picks::as.picks(flag_var_anl),
    fontsize,
    plot_height,
    plot_width,
    transformators
  )

```

averissimo (changes requested):
Awesome! It works great and it even supports `picks` conversion for `choices_selected`

R/tm_g_ae_oview.R
osenan: I can leave as it was. Then I have to modify the lintr configuration, as it was complaining in previous commits

R/tm_g_ae_oview.R
osenan: I can change it, no strong reason

R/tm_g_ae_oview.R
osenan: Done, just had to remove the `pipe_consistency_linter`

R/tm_g_ae_oview.R
osenan: I tried the commit but it does not wok. The reason is, `as.picks` does not support "choices.selected` class.

However, your suggestion is still very valuable, but I need to change the implementation. Which solution do you see more useful:

1. Add a new function to `teal.picks` called `as.variables` that would create a pick object or add a new scenario to `as.picks` that would create a variable for `choices.selected`

2. Add a utils private function in teal.osprey to port "choices.selected" class to `teal.picks::variables`



Although 1 seems better, I wonder after the migration to picks what would be the usefulness of such function, besides not repeating on every module package a util function

averissimo: It does, please pull your `main` branch

```r
teal.picks::as.picks(teal.transform::choices_selected(c("a", "b")))
#>  <variables>
#>    choices: a, b
#>    selected: a
#>    multiple=FALSE, fixed=FALSE, allow-clear=FALSE
```

osenan: True, fixed now

R/tm_g_ae_oview_picks.R
llrs-roche: Yeah, once the module is resolved, it can be accessed directly to retrieve the variables selected (if any). The advantage of using the merged variables is that if they need to be renamed then it will pick the new name (I haven't checked this). 

In this case, I would go with the merged too to avoid any potential issue.

osenan: I did not know that using the merged we can take renamed variables. Sounds better to me as well , thanks

R/tm_g_ae_oview.R
```diff
@@ -73,17 +79,33 @@
...
                           flag_var_anl,
-                          fontsize = c(5, 3, 7),
-                          plot_height = c(600L, 200L, 2000L),
-                          plot_width = NULL,
-                          transformators = list()) {
+                          fontsize,
+                          plot_height,
+                          plot_width,
+                          transformators) {
+  UseMethod("tm_g_ae_oview", arm_var)
```
averissimo: One last change that we're doing in `tmc` where this kind of methods don't require S3 as we can just convert in the function itself and make it even simpler:

https://github.com/insightsengineering/teal.modules.clinical/pull/1476/changes#diff-d50ab07375ab0f048cc866c3f6c0a0e7ed77c779483742bd3eebf05156f2a677

tests/testthat/test-tm_g_ae_oview.R
averissimo, line 1: Awesome addition!

averissimo (changes requested):
Ready to be merged as it works as expected.

I have one last ask that both me and @m7pr have been doing in `tmc` and it also makes sense here.

Instead of `S3` class we can just convert `choices_selected` in the body of the function, in essence support both without the overhead of S3. 

Check this `R/tm_t_pp_basic_info.R`: https://github.com/insightsengineering/teal.modules.clinical/pull/1476/changes#diff-d50ab07375ab0f048cc866c3f6c0a0e7ed77c779483742bd3eebf05156f2a677

R/tm_g_ae_oview.R
osenan: Nice idea! I changed to use a function as we can expect to reuse it across the different modules

R/tm_g_ae_oview.R
```diff
@@ -1,4 +1,4 @@
-#' Teal module for the `AE` overview
+#' @title Teal module for the `AE` overview
```
averissimo: Why change this?

```suggestion
#' Teal module for the `AE` overview
```

R/utils.R
```diff
@@ -239,3 +239,19 @@ set_chunk_dims <- function(pws, q_r, inner_classes = NULL) {
...
 }
+
+#' Wrapper to help converting choices_selected module arguments to picks variables
+#'
+#' @param args Vector with the names of module arguments that can be of class `choices_selected`
+#' @keywords internal
+convert_args_to_picks <- function(args) {
+  checkmate::assert_character(args)
+
+  env <- parent.frame()
```
averissimo: Not a blocker, but consider using `env` in argument as similar methods that change the `parent.frame()` do.

```suggestion
convert_args_to_picks <- function(args, env = parent.frame()) {
  checkmate::assert_character(args)
```

R/tm_g_ae_oview.R
```diff
@@ -72,89 +67,140 @@
...
+  fontsize = c(5, 3, 7),
+  plot_height = c(600L, 200L, 2000L),
+  plot_width = NULL,
+  transformators = list()
+) {
   message("Initializing tm_g_ae_oview")
-  checkmate::assert_class(arm_var, classes = "choices_selected")
-  checkmate::assert_class(flag_var_anl, classes = "choices_selected")
+
+  convert_args_to_picks(c("arm_var", "flag_var_anl"))
```
averissimo: I'm not a fan of having other functions changing the current environment, this breaks the R way of doing things.

do you think the extra code needed per module justifies this approach?

How about if the util function checks if it's a choices_selected and performs `as.picks()` call, otherwise keeps it as is.

```r
arm_var <- .convert_args_to_picks(arm_var)
flag_var_anl <- .convert_args_to_picks(flag_var_anl)
# ...

.convert_arg_to_picks <- function(arg) {
  if (checkmate::test_multiclass(arg, c("choices_selected", "filter_spec", "select_spec")) {
    as.picks(arg)
  } else {
    arg
  }
}

```

R/tm_g_ae_oview.R
```diff
@@ -72,89 +67,140 @@
...
+    selected = 1L
+  ),
+  fontsize = c(5, 3, 7),
+  plot_height = c(600L, 200L, 2000L),
+  plot_width = NULL,
+  transformators = list()
+) {
   message("Initializing tm_g_ae_oview")
-  checkmate::assert_class(arm_var, classes = "choices_selected")
-  checkmate::assert_class(flag_var_anl, classes = "choices_selected")
```
averissimo, line 86: This is no longer present, please keep them with the new argument after the conversion is performed

```r
  # new code conversion of `choices_selected` to `variables`
  # then:

  checkmate::assert_class(arm_var, classes = "variables")
  checkmate::assert_class(flag_var_anl, classes = "variables")

  # but before
  arm_var <- teal.picks::picks(teal.picks::datasets(dataname), arm_var)
  flag_var_anl <- teal.picks::picks(
    teal.picks::datasets(dataname),
    flag_var_anl
  )
```

R/utils.R
```diff
@@ -239,3 +239,19 @@ set_chunk_dims <- function(pws, q_r, inner_classes = NULL) {
...
+  for (arg in args) {
+    val <- get(arg, envir = env)
```
averissimo: Should there be a custom error message in case this `arg` has a typo/doesn't exist?

R/tm_g_ae_oview.R
```diff
@@ -72,89 +67,140 @@
...
+  label,
+  dataname,
+  arm_var = teal.picks::variables(
+    choices = teal.picks::is_categorical(min.len = 2),
+    selected = 1L
+  ),
+  flag_var_anl = teal.picks::variables(
+    choices = teal.picks::is_categorical(min.len = 2),
+    selected = 1L
+  ),
```
averissimo: Do these defaults make sense? As running the app with the defaults makes it hard to find the correct combination.

In `tmc` the examples we're using as defaults were created by statistical programmers, so that makes sense for us to incorporate in the default arguments.

But here I'm not sure if it's better not to have defaults, as the data being used is quite specific

averissimo (changes requested):
It's almost ready! 👍 

My main request is about the default arguments, in this package it might be better to leave them `<missing>`.

The second main comment here is about the implementation of the conversion. I would prefer to keep this only \*if really needed\* as it breaks with best practices of R and this project.

I understand the requirements for the util, but maybe we can find a middle ground, see my suggestion and we can then apply "upstrean" on `tmc`

R/tm_g_ae_oview.R
osenan: I think the defaults make sense if they are useful, ideally with some `teal.data` we could call the module `tm_g_ae_oview()`. In this case, they are not very helpful so I removed them

R/utils.R
osenan: Not needed anymore, I have refactored the function

R/tm_g_ae_oview.R
osenan: I have refactored the part to convert and check before it is picks

R/tm_g_ae_oview.R
osenan: This is an artifact from joining into a single file both methods, no need for change here

R/tm_g_ae_oview.R
osenan: It is better and I refactored using that advice

averissimo (approved):
Awesome! Seems fine by me

R/utils.R
```diff
@@ -239,3 +239,19 @@ set_chunk_dims <- function(pws, q_r, inner_classes = NULL) {
...
+  checkmate::assert(
+    checkmate::check_class(arg, "choices_selected"),
+    checkmate::check_class(arg, "variables")
+  )
+  if (checkmate::test_multi_class(arg, c("choices_selected", "filter_spec", "select_spec"))) {
+    teal.picks::as.picks(arg)
+  } else {
+    arg
+  }
+}
```
averissimo: New candidate code from `tmc`. This allows for better flexbility and verbosity as it names the argument name.

```suggestion
convert_arg_to_picks <- function(arg,
                                 allowed_class = c("choices_selected", "filter_spec", "select_spec"),
                                 envir = parent.frame()) {
  arg <- rlang::ensym(arg)
  arg_name <- as.character(arg)
  arg <- rlang::eval_tidy(arg, env = envir)
  allowed_class <- match.arg(allowed_class)
  checkmate::assert_multi_class(arg, c(allowed_class, "pick", "picks"), .var.name = arg_name)
  teal.picks::as.picks(arg)
}
```

averissimo: (edited)

R/tm_g_ae_oview.R
```diff
@@ -43,27 +44,21 @@
...
 #'   data = data,
 #'   modules = modules(
 #'     tm_g_ae_oview(
 #'       label = "AE Overview",
 #'       dataname = "ADAE",
-#'       arm_var = choices_selected(
-#'         selected = "ACTARM",
-#'         choices = c("ACTARM", "ACTARMCD")
+#'       arm_var = teal.picks::variables(
+#'         choices = dplyr::starts_with("ACTARM"),
```
m7pr: Do not prefix functions in examples nor vignettes.
Only prefix in tests and in the code/implemetnation

R/tm_g_ae_oview.R
```diff
@@ -72,89 +67,133 @@
...
-                          fontsize = c(5, 3, 7),
-                          plot_height = c(600L, 200L, 2000L),
-                          plot_width = NULL,
-                          transformators = list()) {
+#' @export
+tm_g_ae_oview <- function(
+  label,
+  dataname,
+  arm_var,
+  flag_var_anl,
```
m7pr, line 75: I know @averissimo suggested to drop default values, but we do have default values and the checks in tmc :P

R/tm_g_ae_oview.R
```diff
@@ -72,89 +67,133 @@
...
+    attr(arm_var$variables, "multiple") <- FALSE
+  }
+
+  if (isTRUE(attr(flag_var_anl$variables, "multiple"))) {
+    warning(
+      "`flag_var_anl` accepts only a single variable selection. ",
+      "Forcing `teal.picks::variables(multiple)` to FALSE."
+    )
+    attr(flag_var_anl$variables, "multiple") <- FALSE
+  }
```
m7pr: Is this possible to make it a wrapper, and reuse in all modules in this package? You could keep this wrapper in the main feature branch and reuse in single branches for specific modules.

R/utils.R
```diff
@@ -239,3 +239,18 @@ set_chunk_dims <- function(pws, q_r, inner_classes = NULL) {
...
+#' Convert choice selected module variables to picks class
+#'
+#' @param arg module variable of class `choices_selected` or picks variables
+#' @keywords internal
+convert_arg_to_picks <- function(arg,
+                                 allowed_class = c("choices_selected", "filter_spec", "select_spec"),
+                                 envir = parent.frame()) {
+  arg <- rlang::ensym(arg)
+  arg_name <- as.character(arg)
+  arg <- rlang::eval_tidy(arg, env = envir)
```
m7pr: If possible use base R instead of rlang

tests/testthat/test-tm_g_ae_oview.R
```diff
@@ -0,0 +1,69 @@
...
+        label = "AE Overview",
+        dataname = "ADAE",
+        arm_var = list(),
+        flag_var_anl = flag_var_cs
+      ),
+      class = "error"
+    )
+  })
+
+  testthat::it("fails when flag_var_anl is neiter picks or choices_selected", {
```
m7pr: we agreed to add testthat:: prefix for all tests for all functions besides `it` function, as it collideas with some namespace. check other packages in teal family to see that `it` is barely prefixed

DESCRIPTION
```diff
@@ -37,6 +37,7 @@ Imports:
     teal.code (>= 0.7.0),
     teal.data (>= 0.8.0),
     teal.logger (>= 0.4.0),
+    teal.picks (>= 0.1.0),
```
m7pr: teal.transform is in the Depends, and teal.picks is a substitute, so should also be in the Depends. the same we did in tmc

m7pr (changes requested):

R/utils.R
m7pr: Also rlang was not in imports/depends so you get this warning in R CMD CHECK
https://github.com/insightsengineering/teal.osprey/actions/runs/24245023719/job/70788946164?pr=331#step:44:98

R/tm_g_ae_oview.R
m7pr: When you move teal.picks in DESCRIPTION from Imports to Depends, you will not need to prefix those functions for the examples to work. As you didn't need to prefix teal.transform functions

R/tm_g_ae_oview.R
m7pr: this one is not critical

osenan: In this case, the default was empty so I will keep it empty

R/tm_g_ae_oview.R
osenan: Yes, I am using the teal.picks native wrapper

R/utils.R
osenan: finally I am not using rlang

DESCRIPTION
osenan: true, changed

PR #335, opened 2026-03-19
Add picks to tm_g_butterfly module

osenan:
It solves part of #330

Following similar strategy to #331 we are only having one picks srv and ui function and converting `choices_selected` to picks. However, there is a [bug under investigation](https://github.com/insightsengineering/teal.picks/issues/28) on teal.picks and currently the example crashes if we change the value of any variable on the default method.

Please try this example to test the app
```
library(teal)
library(teal.picks)
devtools::load_all()

 data <- teal_data() %>%
   eval_code("set.seed(23) # @linksto ADSL") %>%
   within({
     library(nestcolor)
     library(dplyr)
     ADSL <- rADSL
     ADAE <- rADAE
     ADSL <- mutate(ADSL, DOSE = paste(sample(1:3, n(), replace = TRUE), "UG"))
     ADAE <- mutate(
       ADAE,
       flag1 = ifelse(AETOXGR == 1, 1, 0),
       flag2 = ifelse(AETOXGR == 2, 1, 0),
       flag3 = ifelse(AETOXGR == 3, 1, 0),
       flag1_filt = rep("Y", n())
     )
   })

 app <- init(
   data = data,
   modules = modules(
     tm_g_butterfly(
       label = "Butterfly Plot (picks)",
       dataname = "ADAE",
       right_var = teal.picks::variables(
         choices = c("SEX", "ARM", "RACE"),
         selected = "SEX"
       ),
       left_var = teal.picks::variables(
         choices = c("SEX", "ARM", "RACE"),
         selected = "RACE"
       ),
       category_var = teal.picks::variables(
         choices = c("AEDECOD", "AEBODSYS"),
         selected = "AEBODSYS"
       ),
       color_by_var = teal.picks::variables(
         choices = c("AETOXGR", "None"),
         selected = "AETOXGR"
       ),
       count_by_var = teal.picks::values(
         selected = "# of patients",
         choices = c("# of patients", "# of AEs")
       ),
       facet_var = teal.picks::variables(
         choices = c("RACE", "SEX", "ARM"),
         selected = NULL
       ),
       sort_by_var = teal.picks::values(
         selected = "count",
         choices = c("count", "alphabetical")
       ),
       legend_on = TRUE,
     ),
     tm_g_butterfly(
       label = "Butterfly Plot (default)",
       dataname = "ADAE",
       right_var = choices_selected(
         selected = "SEX",
         choices = c("SEX", "ARM", "RACE")
       ),
       left_var = choices_selected(
         selected = "SEX",
         choices = c("SEX", "ARM", "RACE")
       ),
       category_var = choices_selected(
         selected = "AEBODSYS",
         choices = c("AEDECOD", "AEBODSYS")
       ),
       color_by_var = choices_selected(
         selected = "AETOXGR",
         choices = c("AETOXGR", "None")
       ),
       count_by_var = choices_selected(
         selected = "# of patients",
         choices = c("# of patients", "# of AEs")
       ),
       facet_var = choices_selected(
         selected = NULL,
         choices = c("RACE", "SEX", "ARM")
       ),
       sort_by_var = choices_selected(
         selected = "count",
         choices = c("count", "alphabetical")
       ),
       legend_on = TRUE,
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }

```


osenan:
I have read the CLA Document and I hereby sign the CLA

R/tm_g_butterfly.R
```diff
@@ -332,189 +370,166 @@ srv_g_butterfly <- function(id, data, dataname, label, plot_height, plot_width)
...
+          shiny::need(length(right_var_name) > 0, "Right Dichotomization Variable is required."),
+          shiny::need(length(left_var_name) > 0, "Left Dichotomization Variable is required."),
+          shiny::need(length(category_var_name) > 0, "Category Variable is required."),
+          shiny::need(
+            is.factor(ANL[[right_var_name]]),
+            "Right Dichotomization Variable must be a factor variable, contact developer."
+          ),
+          shiny::need(
+            is.factor(ANL[[left_var_name]]),
+            "Left Dichotomization Variable must be a factor variable, contact developer."
```
osenan: Do you see a problem that it is using `shiny::validate` instead of `teal::validate_input`?

R/tm_g_butterfly.R
```diff
@@ -67,33 +68,33 @@
...
-#'         choices = c("AEDECOD", "AEBODSYS")
+#'       category_var = variables(
+#'         choices = c("AEDECOD", "AEBODSYS"),
+#'         selected = "AEBODSYS"
 #'       ),
-#'       color_by_var = choices_selected(
-#'         selected = "AETOXGR",
-#'         choices = c("AETOXGR", "None")
+#'       color_by_var = variables(
+#'         choices = c("AETOXGR", "None"),
```
llrs-roche: Should None be replaced by allowing no selection? I don't think None is a valid variable on that environment and this doesn't allow to disable the colors.

R/tm_g_butterfly.R
```diff
@@ -125,14 +127,64 @@ tm_g_butterfly <- function(label,
...
-  checkmate::assert_class(facet_var, classes = "choices_selected", null.ok = TRUE)
-  checkmate::assert_class(sort_by_var, classes = "choices_selected")
+
+  right_var <- migrate_choices_selected_to_variables(right_var, multiple = FALSE)
+  left_var <- migrate_choices_selected_to_variables(left_var, multiple = FALSE)
+  category_var <- migrate_choices_selected_to_variables(category_var, multiple = FALSE)
+  color_by_var <- migrate_choices_selected_to_variables(color_by_var, multiple = FALSE)
+  filter_var <- migrate_choices_selected_to_variables(filter_var, null.ok = TRUE)
+  facet_var <- migrate_choices_selected_to_variables(facet_var, null.ok = TRUE)
+  count_by_var <- migrate_choices_selected_to_values(count_by_var)
```
llrs-roche, line 139: This function returns a pick of class values.

R/tm_g_butterfly.R
```diff
@@ -125,14 +127,64 @@ tm_g_butterfly <- function(label,
...
+    facet_var <- create_picks_helper(teal.picks::datasets(dataname), facet_var)
+  }
+
+  if (teal.picks::is_pick_multiple(right_var$variables)) {
+    warning(
+      "`right_var` accepts only a single variable selection. ",
+      "Forcing `teal.picks::variables(multiple)` to FALSE."
+    )
+    attr(right_var$variables, "multiple") <- FALSE
+  }
```
llrs-roche: I think this is also a good candidate for a helper function. Note that is repeated 4 times in only this module. Something like  
```r
as_pick_single <- function(var, name) {
   if (teal.picks::is_pick_multiple(right_var$variables)) {
        warning(
          sprintf("`%s` accepts only a single variable selection. ", name)
          "Forcing `teal.picks::variables(multiple)` to FALSE."
    )
    attr(var "multiple") <- FALSE
    var
}
```

That could be used as `right_var$variables <- as_pick_single(right_var$variables, "right_var")`

R/tm_g_butterfly.R
```diff
@@ -332,189 +370,181 @@ srv_g_butterfly <- function(id, data, dataname, label, plot_height, plot_width)
...
+        left_var_name <- merged$variables()$left_var
+        category_var_name <- merged$variables()$category_var
+        color_by_var_name <- merged$variables()$color_by_var
+        count_by_var_name <- input$count_by_var
+        sort_by_var_name <- input$sort_by_var
+        filter_var_name <- merged$variables()$filter_var
+        facet_var_name <- merged$variables()$facet_var
 
-        teal::validate_inputs(iv())
+        teal::validate_has_data(ANL, min_nrow = 0, msg = "ANL Data is empty")
```
llrs-roche: I guess here we are mimicking previous checks. But, wouldn't it make sense to add `min_nrow = 10` or some other value? This might be a case of leave the code better than what we found.

R/tm_g_butterfly.R
```diff
@@ -332,189 +370,181 @@ srv_g_butterfly <- function(id, data, dataname, label, plot_height, plot_width)
...
+              group_names = c(right_name, left_name),
+              block_count = .(count_by_var_name),
+              block_color = .(if (!is.null(color_by_var_name) && color_by_var_name != "None") {
+                bquote(ANL[[.(color_by_var_name)]])
+              } else {
+                NULL
+              }),
+              id = ANL$USUBJID,
+              facet_rows = .(if (!is.null(facet_var_name)) {
+                bquote(ANL[[.(facet_var_name)]])
```
llrs-roche: Are those many bquotes needed?  Then it might be worth to consider using within 

```suggestion
                ANL[[.(facet_var_name)]]
```

R/utils_picks.R
```diff
@@ -135,6 +135,56 @@ migrate_choices_selected_to_variables <- function(x, # nolint: object_length_lin
...
+  if (inherits(x, "picks")) {
+    return(x)
+  }
+  if (inherits(x, "variables")) {
+    return(x)
+  }
+  if (inherits(x, "choices_selected")) {
+    lifecycle::deprecate_warn(
+      when = "0.5.0",
+      what = I(paste0("`", arg_name, "`")),
```
llrs-roche, line 190: This line seems overcomplicated, it has `paste0()` and `I()`, perhaps something like this:

```suggestion
      what = sQuote(arg_name)
```

or simply `sprintf('`%s`', arg_name)`

llrs-roche (commented):
Great work.

I was a bit surprised when running the app and I saw some input elements weren't using picks for selecting values. The "Choose Up To 2:" can be done with an `teal::validate_input()`, not sure how it has been done on tmc or other packages. 

Image


Very few comments about the code itself. I'd like to have more tests but that is out of the scope for now.

R/tm_g_butterfly.R
osenan: Yes, but I think it is the correct class of the object. It is a `pick` class of values because it is not connected to datasets or variables, the choices are independent of the dataset.

R/tm_g_butterfly.R
osenan: Yes, good observation. I changed to allow no color in the plot and updated the example

R/tm_g_butterfly.R
osenan: I agree with you, but I suggest 1 as minimum for filtering with big constraints

R/tm_g_butterfly.R
osenan: Yes, I changed everything to within so it is more readable

R/utils_picks.R
osenan: This is a sharp comment but unfortunately lifecycle::deprecate_warn requires to use `I` to allow for most of the content, `sprintf` does not work. Originally it has backquotes, single quote using function `sQuote` is slightly different. I need to use additional changes, I prefer to keep it as it is although is not perfect..

R/tm_g_butterfly.R
osenan: Agree! Changed

R/utils_picks.R
llrs-roche: sQuotes can be configured via options. But let's leave it as is to mimic previous behavior.

llrs-roche (approved):

PR #342, opened 2026-07-13
Migrate tm_g_patient_profile to picks

### LINKED ISSUE #330: [Feature Request]: Allow modules to use teal.picks

osenan:
This partially closes #330

This migrates the patient profile module.
It can be tested here:

```
devtools::load_all()
 data <- teal_data() %>%
   within({
     library(nestcolor)
     library(dplyr)
     ADSL <- rADSL
     ADAE <- rADAE %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     ADCM <- rADCM %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     # The step below is to pre-process ADCM to legacy standard
     ADCM <- ADCM %>%
       select(-starts_with("ATC")) %>%
       unique()
     ADRS <- rADRS %>% mutate(ADT = as.Date(ADTM))
     ADEX <- rADEX %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     ADLB <- rADLB %>% mutate(ADT = as.Date(ADTM), LBSTRESN = as.numeric(LBSTRESC))
   })

 ADSL <- data[["ADSL"]]

 app <- init(
   data = data,
   modules = modules(
     tm_g_patient_profile(
       patient_id = variables(
         choices = "USUBJID",
         selected = "USUBJID"
       ),
       sl_dataname = "ADSL",
       ex_dataname = "ADEX",
       ae_dataname = "ADAE",
       rs_dataname = "ADRS",
       cm_dataname = "ADCM",
       lb_dataname = "ADLB",
       sl_start_date = variables(
         selected = "TRTSDTM",
       ),
       ex_var = variables(
         selected = "PARCAT2",
         choices = "PARCAT2"
       ),
       ae_var = variables(
         selected = "AEDECOD",
       ),
       ae_line_col_var = variables(
         selected = "AESER",
         choices = c("AESER", "AEREL")
       ),
       ae_line_col_opt = c("Y" = "red", "N" = "blue"),
       rs_var = variables(
         selected = "PARAMCD",
         choices = "PARAMCD"
       ),
       cm_var = variables(
         selected = "CMDECOD",
         choices = c("CMDECOD", "CMCAT")
       ),
       lb_var = variables(
         selected = "LBTESTCD",
         choices = c("LBTESTCD", "LBCAT")
       ),
       x_limit = "-28, 750",
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }

```
I had to use teal.picks values for the patient id.
Please check that CI passes the new tests.


R/tm_g_patient_profile.R
```diff
@@ -459,20 +475,130 @@ srv_g_patient_profile <- function(id,
...
+        )
+        teal::validate_input(
+          "sl_start_date-variables-selected",
+          condition = function(x) length(x) > 0L,
+          message = "Date variable is required."
+        )
+        if (isTRUE(select_plot()[ex_dataname])) {
+          teal::validate_input(
+            "ex_var-variables-selected",
+            condition = function(x) length(x) > 0L,
```
llrs-roche: Not sure how this is triggering the validation error: the UI shows one variable and it should run.

R/tm_g_patient_profile.R
```diff
@@ -238,82 +293,71 @@ ui_g_patient_profile <- function(id, ...) {
             )
           )
         ),
```
llrs-roche, line 305: This could also be a `teal.picks::datasets()` selection. Just pointing it out because it might facilitate downstream processing.

R/tm_g_patient_profile.R
```diff
@@ -356,23 +423,35 @@ srv_g_patient_profile <- function(id,
   checkboxes <- c(ex_dataname, ae_dataname, rs_dataname, lb_dataname, cm_dataname)
   moduleServer(id, function(input, output, session) {
     teal.logger::log_shiny_input_changes(input, namespace = "teal.osprey")
-    select_plot <- reactive(
-      vapply(checkboxes, function(x) x %in% input$select_ADaM, logical(1L))
-    )
 
-    resolved <- teal.transform::resolve_delayed(patient_id, as.list(isolate(data())))
+    selectors <- teal.picks::picks_srv(
+      id = "",
```
llrs-roche: I think it would be better to have an id. It would make it easier to call and validate.

R/tm_g_patient_profile.R
```diff
@@ -459,20 +475,130 @@ srv_g_patient_profile <- function(id,
...
+        ae_var <- pick_selected("ae_var")
+        ae_line_col_var <- pick_selected("ae_line_col_var")
+        rs_var <- pick_selected("rs_var")
+        cm_var <- pick_selected("cm_var")
+        ex_var <- pick_selected("ex_var")
+        lb_var <- pick_selected("lb_var")
         x_limit <- input$x_limit
         lb_var_show <- input$lb_var_show
 
+        validated_q <- obj
```
llrs-roche: Why do we need this renaming? `validated_q` is used downstream but no usage of `obj`.

R/tm_g_patient_profile.R
```diff
@@ -324,18 +368,41 @@ ui_g_patient_profile <- function(id, ...) {
...
   )
 }
 
+#' @keywords internal
+.patient_profile_picks_selected_var <- function(selector_state) {
+  if (is.null(selector_state) || is.null(selector_state$variables)) {
+    return(character())
+  }
+  as.character(selector_state$variables$selected)
+}
```
llrs-roche: I don't think these functions (with `.patient_profile_picks_selected_values()`) are needed. How have the other modules work around this issue? Besides they are general enough that the name could be `picks_selected_values()`.

R/tm_g_patient_profile.R
```diff
@@ -459,20 +475,130 @@ srv_g_patient_profile <- function(id,
           )
         obj <- teal.code::eval_code(obj, "library(dplyr)")
 
-        teal::validate_inputs(iv())
+        pick_selected <- function(name) {
```
llrs-roche: What's wrong with `selectors$rs_var()$variables$selected` and such calls used in other modules?

llrs-roche (commented):
The example app requires a variable that is there but locked and it doesn't the server to render the output:

Image

I think there is some room for simplification on the logic and code. There is more addition than shortening: +503 -241.

There are some warnings on tests:

```
Warning (test-tm_g_patient_profile.R:25:9): tm_g_patient_profile input validation / plot arguments input validation
`patient_id` was deprecated in teal.osprey 0.5.0.
ℹ Pass `teal.picks::variables()` (or a full `teal.picks::picks()` chain). Support for
  legacy `teal.transform::choices_selected()`, `filter_spec`, and `select_spec` is
  deprecated.
Backtrace:
    ▆
 1. ├─testthat::expect_error(...) at test-tm_g_patient_profile.R:23:5
 2. │ └─testthat:::quasi_capture(...)
 3. │   ├─testthat (local) .capture(...)
 4. │   │ └─base::withCallingHandlers(...)
 5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 6. ├─base::suppressWarnings(...) at test-tm_g_patient_profile.R:25:9
 7. │ └─base::withCallingHandlers(...)
 8. └─teal.osprey::tm_g_patient_profile(...)
 9.   └─teal.osprey:::migrate_choices_selected_to_variables(...) at teal.osprey/R/tm_g_patient_profile.R:187:3

Warning (test-tm_g_patient_profile.R:25:9): tm_g_patient_profile input validation / plot arguments input validation
`sl_start_date` was deprecated in teal.osprey 0.5.0.
ℹ Pass `teal.picks::variables()` (or a full `teal.picks::picks()` chain). Support for
  legacy `teal.transform::choices_selected()`, `filter_spec`, and `select_spec` is
  deprecated.
Backtrace:
    ▆
 1. ├─testthat::expect_error(...) at test-tm_g_patient_profile.R:23:5
 2. │ └─testthat:::quasi_capture(...)
 3. │   ├─testthat (local) .capture(...)
 4. │   │ └─base::withCallingHandlers(...)
 5. │   └─rlang::eval_bare(quo_get_expr(.quo), quo_get_env(.quo))
 6. ├─base::suppressWarnings(...) at test-tm_g_patient_profile.R:25:9
 7. │ └─base::withCallingHandlers(...)
 8. └─teal.osprey::tm_g_patient_profile(...)
 9.   └─teal.osprey:::migrate_choices_selected_to_variables(...) at teal.osprey/R/tm_g_patient_profile.R:188:3
```

R/tm_g_patient_profile.R
osenan: The problem is that if a picks input is `fixed` part of the input is hidden and it is not seen in the validation. I have added a utils function to validate pick inputs that are fixed.

R/tm_g_patient_profile.R
osenan: It is a pattern that we were using in tmc to facilitate the identification of qenv dedicated solely or mostly to input validation, but maybe here is unnecessary

R/tm_g_patient_profile.R
osenan: No, apologies this was an hallucination from LLM that should not have been included into the PR. I have fixed and removed them.

R/tm_g_patient_profile.R
osenan: The only advantage is for intermediate results if the value does not exist it will be NULL. While empty character will not create issues will does create. I have tried removing the function but simply using the selector create an error.

R/tm_g_patient_profile.R
osenan: Changed

R/tm_g_patient_profile.R
osenan: I think so, it will probably simplify and remove code that is currently used in the server to create the choices and select the values. It has additional checks for the empty datasets,  that is the part which is not straightforward to me. I suggest creating an issue as it will require some time.

R/tm_g_patient_profile.R
llrs-roche: If this is a pattern across the packages it is fine. I have just missed it so far when I looked on the tmc code. 
I though that simply instead of `obj` it could directly create a `validated_q` as there are no changes on `obj`.

R/tm_g_patient_profile.R
llrs-roche: Feel free to open an issue, as I'm not sure what is the problem with checks for empty output from teal.picks is.

R/tm_g_patient_profile.R
llrs-roche: I am not sure I'm following you. 
1) Selectors in other modules are always assumed they might need to be merged, and here we don't (I think this deals with the case of missing selectors).
2) If they are NULL they can be skipped as it is done in other parts of the module.

I think we should also check that apps with default arguments work. Which isn't the case. I think we need to check it well. Even if I add the minimal required values the app starts but fails:

```
app <- init(
  data = data,
  modules = modules(
    tm_g_patient_profile(label = "Patient Profile Plot",
                         patient_id = variables(
                           choices = unique(ADSL$USUBJID),
                           selected = unique(ADSL$USUBJID)[1]
                         ),
                         sl_dataname = "ADSL",
                         ex_dataname = "ADEX",
                         ae_dataname = "ADAE",
                         rs_dataname = "ADRS",
                         cm_dataname = "ADCM",
                         lb_dataname = "ADLB",
                         sl_start_date = variables(
                                   selected = "TRTSDTM",
                                 )
    )
  )) |> runApp()
```

image

The *_dataname variables can be a selection of datasets that start with AD and end with the respective letters (or something along those lines).

R/tm_g_patient_profile.R
osenan: I think we do not need `validated_q` in this module. It is correct without it

R/tm_g_patient_profile.R
osenan: I tried but `teal.picks::datasets` has a problem in its design. It is constrained so datasets with non fixed choices only have available single choice.
`teal.picks::datasets(choices = c("dataset1", "dataset2"), multiple = TRUE)` fails because it is not expected that user will select variables from different datasets.
We could change teal.picks but I found it too much for an improvement on this module.

R/tm_g_patient_profile.R
osenan: Solved simply changing the function to validate

R/tm_g_patient_profile.R
osenan: I have done the two changes.
1. Remove the `picks_selector` function. It was actually not required. Thanks!
2. Add defaults where it was possible. There are lots of harcoding and inner filtering/data processing into this module. The variables that do not have default is because it is not useful to have it one. At least I have changed the defaults, because no argument can actually be NULL.

osenan: I have actually reverted the changes. Because some variables that are picks can be NULL, we need a helper function that returns character() if the pick is set to NULL. If I directly use the selector the reactive object `selectors$my_NULL_picks()` does not exist and throws an error.
The default need to be NULL, as some plots and variables are optional.

llrs-roche (changes requested):
Try setting up an app with just this module without any argument that it is optional:

```r
app <- init(
  data = data,
  modules = modules(
    tm_g_patient_profile(
      patient_id = variables(
        choices = USUBJID,
        selected = USUBJID
      ),
      sl_start_date = variables(
        selected = "TRTSDTM",
      ),
      sl_dataname = "ADSL"
    )
  )
) |> runApp()
```

You'll see an app (using the same data as per last comment) will show this validation:

image

That's good, but we **also** want that the default module generates a working output for the user, so that teal app developers don't need to worry too much when they deploy new apps.  

We also talked today about not using functions on `validate_input()`'s condition argument. This will need to be done here or later with #340, so please let's start here with the modules that are not yet finished. 

I also mentioned that the list of checkboxes for `*_dataname` could be improved with teal.picks and I think this is not a scope creep. Instead of multiple inputs and checkboxes, users could get one input (or one input per module)

R/tm_g_patient_profile.R
llrs-roche: In other modules the selectors that are null are filtered out with `Filter(Negate(is.null)), selectors)`, something like this would be easy to implement without a helper.

R/tm_g_patient_profile.R
llrs-roche: I agree it might be too much for just this module. But I think the `datasets()` should be more flexible. I've opened an issue there to discuss it: https://github.com/insightsengineering/teal.picks/issues/107

R/tm_g_waterfall.R
```diff
@@ -515,12 +517,14 @@ srv_g_waterfall <- function(id,
         })
       )
 
+      rs_sub <- NULL # To avoid no visible binding R CMD Check NOTE
```
llrs-roche: Minor comment, but I don't see that NOTE on the CI for 340 for example: https://github.com/insightsengineering/teal.osprey/actions/runs/30093724791/job/89482684511?pr=344

R/tm_g_waterfall.R
```diff
@@ -515,12 +517,14 @@ srv_g_waterfall <- function(id,
...
+
       q1 <- if (is.null(adrs_paramcd)) {
         teal.code::eval_code(
           q1,
           code = bquote({
             anl <- bar_data
-            anl$USUBJID <- unlist(lapply(strsplit(anl$USUBJID, "-", fixed = TRUE), tail, 1))
+            anl$USUBJID <- unlist(lapply(strsplit(anl$USUBJID, "-", fixed = TRUE), utils::tail, 1))
```
llrs-roche: Why is the prefix needed?? It is base and loaded by default. It would only fail if the user changed the automatically loaded packages not to include utils.

R/tm_g_patient_profile.R
```diff
@@ -168,13 +183,56 @@ tm_g_patient_profile <- function(label = "Patient Profile Plot",
...
+  if (!is.null(lb_var) && !is.na(lb_dataname)) {
+    lb_var <- create_picks_helper(teal.picks::datasets(lb_dataname), lb_var)
+  }
+
+  check_variable_and_dataset(ex_var, ex_dataname)
+  check_variable_and_dataset(ae_var, ae_dataname)
+  check_variable_and_dataset(ae_line_col_var, ae_dataname)
+  check_variable_and_dataset(rs_var, rs_dataname)
+  check_variable_and_dataset(lb_var, lb_dataname)
+  check_variable_and_dataset(cm_var, cm_dataname)
```
llrs-roche: This new function should either use assertions and then I can make it use on other modules or be removed.

llrs-roche (approved):
Great addition with the tests.

I think the `check_variable_and_dataset()` should either not exists or use assertions so it can be used on `tm_g_heat_bygrade()` too. 

Just two other minor comments.
Thanks and sorry for the long review!

R/tm_g_waterfall.R
osenan: It failed to me locally and I added the checks. But in CI they do not appear. It does not make sense as other modules also use `tail` and the CI NOTE does not appear.

R/tm_g_waterfall.R
osenan: Yes, I observed locally and added the change. But in the CI it does not appear.

PR #341, opened 2026-07-09
Migrate tm_g_terms_id to picks

osenan:
# Pull Request

It partially fixes #330 

Please try the module with this example:

```
devtools::load_all()
 data <- teal_data() %>%
   within({
     ADSL <- rADSL
     ADAE <- rADAE
   })

 join_keys(data) <- default_cdisc_join_keys[names(data)]

 app <- init(
   data = data,
   modules = modules(
     tm_g_events_term_id(
       label = "Common AE",
       term_dataname = "ADAE",
       term_var = variables(
         choices = is_categorical(min.len = 2),
         selected = "AEDECOD"
       ),
       arm_dataname = "ADSL",
       arm_var = variables(
         choices = is_categorical(min.len = 2),
         selected = "ACTARMCD"
       ),
       plot_height = c(600, 200, 2000)
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }

```

This PR is replacing #329

PR #343, opened 2026-07-14
Simpify design of tm_g_events_term_id, fix failing CI fixes

osenan:
# Pull Request

This is part of #330 

This is replacing PR #341. The module `tm_g_term_id` was already on main. Altough the implementation worked, in my opinion it had too complex code which was not required. I have tried to simplify the migration starting from scratch. Then I also made modifications on input validation and use the `within.qenv` method to have code easier to read.

In this same PR I tried to fix all the errors on CI. Hopefully when this will be merged we can merge it with the other PRs and have clean checks.

The module can be tested with this example:

```
devtools::load_all()
 data <- teal_data() %>%
   within({
     ADSL <- rADSL
     ADAE <- rADAE
   })

 app <- init(
   data = data,
   modules = modules(
     tm_g_events_term_id(
       label = "Common AE",
       dataname = "ADAE",
       term_var = variables(
         choices = c(
           "AEDECOD", "AETERM",
           "AEHLT", "AELLT", "AEBODSYS"
         ),
         selected = "AEDECOD"
       ),
       arm_var = variables(
         choices = c("ACTARM", "ACTARMCD"),
         selected = "ACTARMCD"
       ),
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }

``
I found that changing control and treatment values, as changing variables many times does not produce a plot. In addition, the input validation is not very informative. Unfortunately, and due to time restriction. I could not find a pattern that allowed me to improve the input validation.

I checked that in main there is the same behavior related with treatment/control input selection and generation or not of plots. Is something that I will create an issue to improve it.

### CODE REVIEW

File: R/tm_g_events_term_id.R [hunk 1]
```diff
@@ -355,51 +337,84 @@ srv_g_events_term_id <- function(id,
...
-            sort_by = .(input$sort),
-            rate_range = .(input$raterange),
-            diff_range = .(input$diffrange),
-            reversed = .(input$reverse),
-            conf_level = .(input$conf_level),
-            diff_ci_method = .(input$diff_ci_method),
-            axis_side = .(input$axis),
-            fontsize = .(font_size()),
+            arm = ANL[[arm_var_name]],
+            arm_N = table(get(arm_dataset)[[arm_var_orig]]),
```
osenan: I tried to avoid using `get` but it was not working. What I tried is to pass directly the dataset. I think that, because inputs are connected, if you remove the `get` with the lazy evaluation of teal.picks and qenv the plot is never created.

R/tm_g_events_term_id.R
```diff
@@ -33,19 +33,17 @@
...
 #'     tm_g_events_term_id(
 #'       label = "Common AE",
-#'       term_var = teal.picks::picks(
-#'         teal.picks::datasets("ADAE"),
-#'         teal.picks::variables(
-#'           choices = teal.picks::is_categorical(min.len = 2),
-#'           selected = "AEDECOD"
-#'         )
+#'       dataname = "ADAE",
+#'       term_var = variables(
```
llrs-roche, line 40: Should `variables` also be prefixed with  `teal.picks::` or it is imported?

R/tm_g_events_term_id.R
```diff
@@ -55,37 +53,26 @@
...
-  checkmate::assert_class(arm_var, "picks", .var.name = "arm_var")
-  checkmate::assert_false(
-    teal.picks::is_pick_multiple(arm_var$variables),
-    .var.name = "`arm_var` must use variables(..., multiple = FALSE)"
-  )
+  term_var <- migrate_choices_selected_to_variables(term_var)
+  arm_var <- migrate_choices_selected_to_variables(arm_var)
+
+  term_var <- create_picks_helper(teal.picks::datasets(dataname, dataname), term_var)
+  arm_var <- create_picks_helper(teal.picks::datasets("ADSL", "ADSL"), arm_var)
```
llrs-roche: Why we need this dataset instead of whatever is passed by the user from `dataname`?

R/tm_g_events_term_id.R
```diff
@@ -97,33 +84,29 @@ tm_g_events_term_id <- function(label = "Common AE",
...
+    .var.name = "plot_width"
   )
 
-  pick_slots <- list(term_var = term_var, arm_var = arm_var)
-  all_datanames <- .picks_all_datanames(pick_slots)
-
   args <- as.list(environment())
 
   module(
+    datanames = c("ADSL", dataname),
```
llrs-roche: Not sure if this is checked but if the input is `ADSL` we might not want to have a long vector: 

```suggestion
    datanames = unique(c("ADSL", dataname)),
```

R/tm_g_events_term_id.R
```diff
@@ -355,51 +337,84 @@ srv_g_events_term_id <- function(id,
...
-        ANL,
+      teal::validate_has_data(ANL,
         min_nrow = 10,
         msg = "Analysis data set must have at least 10 data points"
       )
 
       teal.reporter::teal_card(qenv) <- c(teal.reporter::teal_card(qenv), "### Plot")
 
-      teal.code::eval_code(
+      # This is redundant only added to avoid R CMD NOTES
```
llrs-roche: Which note? If it is undeclared variable I would recommend to state so.
Also I tend to just declare them with NULL:

```suggestion
      # This is redundant only added to avoid R CMD NOTE on no visible binding: "
      arm_ref <-  arm_trt <- sort_by <- rate_range <- diff_range <- reversed <- conf_level <-  diff_ci_method <- axis_side <- fontsize <- NULL
```

R/tm_g_events_term_id.R
llrs-roche: I'll play a bit with it and see what I come up with.

R/tm_g_events_term_id.R
osenan: It is imported. In tmc we agreed that in examples we do not use prefix, I am following the same here.

R/tm_g_events_term_id.R
osenan: I cannot answer the question completely. Before the migration, `arm_var` was also hardcoded to "ADSL". In development I tried assigning to `dataname` but the plot does not display:
image
There is a table calculation that maybe needs to be done only with this dataset, the one that also contains the variables.

averissimo: Maybe we could add a new parameter called `parent_dataname` which defaults to `ADSL` as other models have

WDYT?

R/tm_g_events_term_id.R
```diff
@@ -97,33 +84,29 @@ tm_g_events_term_id <- function(label = "Common AE",
...
+    .var.name = "plot_width"
   )
 
-  pick_slots <- list(term_var = term_var, arm_var = arm_var)
-  all_datanames <- .picks_all_datanames(pick_slots)
-
   args <- as.list(environment())
 
   module(
+    datanames = unique(c("ADSL", dataname)),
```
averissimo: We might need to copy `.picks_dataname` from `tmg` here as use it instead

Just in case people use a fully specified picks instead of just relying on `dataname`

```r
.picks_datanames <- function(x) {
  checkmate::assert_list(x, c("picks", "NULL"))
  datanames_list <- lapply(x, function(x) {
    if (is.character(x$datasets$choices)) {
      x$datasets$choices
    } else {
      NULL
    }
  })

  if (any(vapply(datanames_list, is.null, logical(1)))) {
    "all"
  } else {
    unique(unlist(datanames_list))
  }
}
```

R/tm_g_events_term_id.R
```diff
@@ -219,21 +200,21 @@ ui_g_events_term_id <- function(id,
...
                                  data,
+                                 dataname,
+                                 label,
                                  term_var,
                                  arm_var,
                                  plot_height,
                                  plot_width) {
   checkmate::assert_class(data, "reactive")
-  checkmate::assert_class(isolate(data()), "teal_data")
+  checkmate::assert_class(shiny::isolate(data()), "teal_data")
```
averissimo: Why add the prefix? it seems out of place with rest of function/package

R/tm_g_events_term_id.R
averissimo: The diff below does the trick, basically adding substitute with `as.name(arm_dataset)` that translates to `ADSL`.

```diff
diff --git a/R/tm_g_events_term_id.R b/R/tm_g_events_term_id.R
index db0db545..abb4adfa 100644
--- a/R/tm_g_events_term_id.R
+++ b/R/tm_g_events_term_id.R
@@ -375,7 +376,7 @@ srv_g_events_term_id <- function(id,
             term = ANL[[term_var_name]],
             id = ANL$USUBJID,
             arm = ANL[[arm_var_name]],
-            arm_N = table(get(arm_dataset)[[arm_var_orig]]),
+            arm_N = table(arm_dataset[[arm_var_orig]]),
             ref = arm_ref,
             trt = arm_trt,
             sort_by = sort_by,
@@ -392,6 +393,7 @@ srv_g_events_term_id <- function(id,
         term_var_name = term_var_name,
         arm_var_name = arm_var_name,
         arm_dataset = arm_dataset,
+        arm_dataset_name = as.name("arm_dataset"),
         arm_var_orig = arm_var_orig,
         arm_ref = input$arm_ref,
         arm_trt = input$arm_trt,

```

R/tm_g_events_term_id.R
osenan: I agree, I will added then the module is more flexible

R/tm_g_events_term_id.R
osenan: Agreed, I updated and added a checkpoint to verify if other modules require of [this util function](https://github.com/insightsengineering/teal.osprey/issues/340)

R/tm_g_events_term_id.R
osenan: Yes, it is wrong, removed

R/tm_g_events_term_id.R
osenan: It works, thanks!

averissimo (approved):
Should be good to go. We still have some things to do (re-write swimlane and waterfall), but let's merge this

PR #336, opened 2026-03-24
Use picks in tm_g_spiderplot module

osenan:
Partially closes issue #330 

I also found a bug while testing the examples, [if we have only one choice the UI of the picks component does not look ok](https://github.com/insightsengineering/teal.picks/issues/33).

The changes are:

Create method dispatch based on `x_var`. Refactor ui and server function to use picks instead of choices selected. I was able to remove some code of features handled by picks automatically like the merge. Maybe part of the validation should be implemented again, but I think I could remove it as well.

This app is a good test, it assumes you have `teal.picks` installed from latest main and are in this branch.

```
library("teal")
library("teal.picks")
devtools::load_all()

data <- teal_data() %>%
  within({
    library(nestcolor)
    ADSL <- teal.data::rADSL
    ADTR <- teal.data::rADTR
  })

join_keys(data) <- default_cdisc_join_keys[names(data)]

app <- init(
  data = data,
  modules = modules(
    tm_g_spiderplot(
      label = "Spider plot (picks)",
      dataname = "ADTR",
      paramcd = teal.picks::variables(
        choices = "PARAMCD",
        selected = "PARAMCD"
      ),
      x_var = teal.picks::variables(
        choices = dplyr::where(is.numeric),
        selected = 1L
      ),
      y_var = teal.picks::variables(
        choices = c("PCHG", "CHG", "AVAL"),
        selected = "PCHG"
      ),
      marker_var = teal.picks::variables(
        choices = c("SEX", "RACE", "USUBJID"),
        selected = "SEX"
      ),
      line_colorby_var = teal.picks::variables(
        choices = c("SEX", "USUBJID", "RACE"),
        selected = "SEX"
      ),
      xfacet_var = teal.picks::variables(
        choices = c("SEX", "ARM"),
        selected = "SEX"
      ),
      yfacet_var = teal.picks::variables(
        choices = c("SEX", "ARM"),
        selected = "ARM"
      ),
      vref_line = "10, 37",
      href_line = "-20, 0"
    ),
    tm_g_spiderplot(
      label = "Spider plot (default)",
      dataname = "ADTR",
      paramcd = choices_selected(
        choices = "PARAMCD",
        selected = "PARAMCD"
      ),
      x_var = choices_selected(
        choices = c("ADY", "AGE"),
        selected = "ADY"
      ),
      y_var = choices_selected(
        choices = c("PCHG", "CHG", "AVAL"),
        selected = "PCHG"
      ),
      marker_var = choices_selected(
[...]

PR #344, opened 2026-07-20
teal.picks review for consistency across modules & packages

llrs-roche:
# Pull Request

Fixes #340 after #330

Please verify the following list:

## Documentation

- [ ] All migrated module argument documentation follow the same documentation pattern if they are equivalent
- [x] Update argument_convention.R file, check for exceptions in modules.: 20th July  @llrs-roche Partially
- [x] Examples do not use prefix for teal.picks objects: 20th July  @llrs-roche 

## Code

- [x] Input validation is implemented (when possible) using teal validate functions: 20th July @llrs-roche; there are some modules that need to be merged and some that will be rewritten => Partially fixed
- [ ] Big chunks of input validation will have their own separated qenv `validated_q` when possible
- [x] We use package helper functions to convert from teal.transform and create pick: 24st July  @llrs-roche; only swimlane, and patient_profile  don't use helpers when they should.
- [x] We avoid user to define picks and try to use teal.picks::variables or teal.picks::picks without dataset: 24th July  @llrs-roche; only swimlane, and patient_profile  don't use helpers when they should.
- [x] Consistent usage of utils function to check and force specific picks not to be multiple: 24st July  @llrs-roche; only swimlane, and patient_profile  don't use helpers when they should.
- [ ] Outputs are created using within and not using teal.code::eval_code: 21st July  @llrs-roche; While many packages use eval_code() not just for setting the required libraries. I don't think this is a high priority, delaying it for now
- [x] Default argument value is consistent across modules
   24st July  @llrs-roche; Removed several lines on examples that matched the definition. There are two modules with different argument values:
 ```
 R/tm_g_butterfly.R:121:                           legend_on = TRUE,
 R/tm_g_spiderplot.R:99:                            legend_on = FALSE,
 ```
 They are mimicking the defaults on `osprey::g_butterfly()` and `osprey::g_spiderplot()`
  There's also the `ui_g_decorate()` that has a different setting for fontsize. I think this makes sense as they are for titles and footnotes. 
  ```
  R/tm_g_ae_oview.R:74:  fontsize = c(5, 3, 7),
  R/tm_g_ae_sub.R:59:                        fontsize = c(5, 3, 7),
  R/tm_g_decorate.R:16:                          fontsize = c(5, 4, 11)) {
  R/tm_g_events_term_id.R:64:                                fontsize = c(5, 3, 7),
  R/tm_g_heat_bygrade.R:136:  fontsize = c(5, 3, 7),
  ```
  `anno_txt_var` is also different for swimlane and spiderplot but that needs to be revised after the update on swimlane module
  ```
  R/tm_g_spiderplot.R:98:                            anno_txt_var = TRUE,
  R/tm_g_swimlane.R:53:                          anno_txt_var = NULL,
  ```

  I think there might be room for some improvements with arguments as for example on waterfall there is a `bar_paramcd`  (with no default) while in other modules there is a just a `paramcd`. 

## Tests

- [x] All tests are passed
- [x] Each module at least has a testServer so we verify that it can be created using teal.picks or teal.transform objects

## Optional

- [ ] Rename differently [variables that replace arguments with the same name](https://github.com/insightsengineering/teal.osprey/pull/336#discussion_r3576955361) to facilitate reading.
- [ ] Check if modules that use `datanames` [require to use utils function from teal.modules.general to extract datanames](https://github.com/insightsengineering/teal.osprey/pull/343#discussion_r3587615766) if user decides to pass full `teal.picks::picks` to module instead of `teal.picks::variables`.


osenan (approved):
I have put a summary comment of all the checks.
My recomendation is to fix all CI issues except tests failing and we can merge.

PR #345, opened 2026-07-20
Migrate `tm_g_waterfall` to `teal.picks`

averissimo:
# Pull Request

- part of #330

### Notable changes:

- New functionality in `migrate_value_choices_to_picks` as this module was very simplistic.
  - Now supports setting only `values` as before and assume the `PARAMCD` variable
- `parentname` is now a parameter of the module (was assumed to be `ADSL`

### Example

```r
# example waterfall
pkgload::load_all()
data <- teal_data() %>%
  within({
    library(nestcolor)
    ADSL <- rADSL
    ADRS <- rADRS
    ADTR <- rADTR
    ADSL$SEX <- factor(ADSL$SEX, levels = unique(ADSL$SEX))
  })
#'
#'
app <- init(
  data = data,
  modules = modules(
    tm_g_waterfall(
      label = "Waterfall",
      dataname_tr = "ADTR",
      dataname_rs = "ADRS",
      bar_paramcd = values(c("SLDINV"), "SLDINV", multiple = FALSE),
      bar_var = variables(c("PCHG", "AVAL"), "PCHG"),
      bar_color_var = variables(c("ARMCD", "SEX"), "ARMCD"),
      bar_color_opt = NULL,
      sort_var = variables(c("ARMCD", "SEX"), NULL),
      add_label_var_sl = variables(c("SEX", "EOSDY"), NULL),
      add_label_paramcd_rs = values(c("BESRSPI", "OBJRSPI"), NULL, multiple = TRUE),
      anno_txt_var_sl = variables(c("SEX", "ARMCD", "BMK1", "BMK2"), NULL, multiple = TRUE),
      anno_txt_paramcd_rs = values(c("BESRSPI", "OBJRSPI"), NULL),
      facet_var = variables(c("SEX", "ARMCD", "STRATA1", "STRATA2"), NULL),
      href_line = "-30, 20"
    )
  )
)
if (interactive()) {
  shinyApp(app$ui, app$server)
}
```

### DIFF
--- a/R/tm_g_waterfall.R
+++ b/R/tm_g_waterfall.R
@@ -280,69 +280,6 @@ ui_g_waterfall <- function(id,
         teal.picks::picks_ui(ns("facet_var"), facet_var),
         helpText("from ", tags$code(parentname))
       ),
-
-      # teal.widgets::optionalSelectInput(
-      #   ns("bar_paramcd"),
-      #   "Tumor Burden Parameter",
-      #   multiple = FALSE
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("bar_var"),
-      #   "Bar Height",
-      #   choices = get_choices(a$bar_var$choices),
-      #   selected = a$bar_var$selected,
-      #   multiple = FALSE,
-      #   label_help = helpText("Tumor change variable from ", tags$code("ADTR"))
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("bar_color_var"),
-      #   "Bar Color",
-      #   choices = get_choices(a$bar_color_var$choices),
-      #   selected = a$bar_color_var$selected,
-      #   multiple = FALSE
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("sort_var"),
-      #   "Sort by",
-      #   choices = get_choices(a$sort_var$choices),
-      #   selected = a$sort_var$selected,
-      #   multiple = FALSE,
-      #   label_help = helpText("from ", tags$code("ADSL"))
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("add_label_var_sl"),
-      #   "Add ADSL Label to Bars",
-      #   choices = get_choices(a$add_label_var_sl$choices),
-      #   selected = a$add_label_var_sl$selected,
-      #   multiple = FALSE
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("add_label_paramcd_rs"),
-      #   "Add ADRS Label to Bars",
-      #   multiple = FALSE
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("anno_txt_var_sl"),
-      #   "Annotation Variables",
-      #   choices = get_choices(a$anno_txt_var_sl$choices),
-      #   selected = a$anno_txt_var_sl$selected,
-      #   multiple = TRUE,
-      #   label_help = helpText("from ", tags$code("ADSL"))
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("anno_txt_paramcd_rs"),
-      #   "Annotation Parameters",
-      #   multiple = TRUE,
-      #   label_help = helpText("from ", tags$code("ADRS"))
-      # ),
-      # teal.widgets::optionalSelectInput(
-      #   ns("facet_var"),
-      #   "Facet by",
-      #   choices = get_choices(a$facet_var$choices),
-      #   selected = NULL,
-      #   multiple = FALSE,
-      #   label_help = helpText("from ", tags$code("ADSL"))
-      # ),
       checkboxInput(
         ns("show_value"),
         "Add Bar Height Value",
--- a/R/tm_g_waterfall.R
+++ b/R/tm_g_waterfall.R
@@ -13,51 +13,51 @@
 #' be available in the list passed to the `data` argument of [teal::init()]
 #' @param dataname_rs (`character(1)`) response analysis data used in teal module to label response parameters, needs to
 #' be available in the list passed to the `data` argument of [teal::init()]
-#' @param bar_paramcd ([teal.picks::values()] or `choices_selected`) parameter in tumor burden data that will be plotted as
-#' bar height.
+#' @param bar_paramcd ([teal.picks::values()] or `choices_selected`)
+#' parameter in tumor burden data that will be plotted as bar height.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::values()] instead.
-#' @param bar_var ([teal.picks::variables()] or [teal.transform::choices_selected()]) numeric variable from dataset to plot the bar height,
-#' e.g., `PCHG`.
+#' @param bar_var ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' numeric variable from dataset to plot the bar height, e.g., `PCHG`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
-#' @param bar_color_var ([teal.picks::variables()] or [teal.transform::choices_selected()]) color by variable (subject level), `None` corresponds
-#' to `NULL`.
+#' @param bar_color_var ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' color by variable (subject level), `None` corresponds to `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
 #' @param bar_color_opt aesthetic values to map color values (named vector to map color values to each name).
 #' If not `NULL`, please make sure this contains all possible values for `bar_color_var` values,
 #' otherwise color will be assigned by `ggplot` default, please note that `NULL` needs to be specified
 #' in this case
-#' @param sort_var ([teal.picks::variables()] or [teal.transform::choices_selected()]) sort by variable (subject level), `None` corresponds
-#' to `NULL`.
+#' @param sort_var ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' sort by variable (subject level), `None` corresponds to `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
-#' @param add_label_var_sl ([teal.picks::variables()] or [teal.transform::choices_selected()]) add label to bars (subject level), `None`
-#' corresponds to `NULL`.
+#' @param add_label_var_sl ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' add label to bars (subject level), `None` corresponds to `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
-#' @param add_label_paramcd_rs ([teal.picks::values()] or [teal.transform::choices_selected()]) add label to bars (response dataset), `None`
-#' corresponds to `NULL`. At least one of `add_label_var_sl` and `add_label_paramcd_rs` needs
-#' to be `NULL`.
+#' @param add_label_paramcd_rs ([teal.picks::values()] or [teal.transform::choices_selected()])
+#' add label to bars (response dataset), `None` corresponds to `NULL`.
+#' At least one of `add_label_var_sl` and `add_label_paramcd_rs` needs to be `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::values()] instead.
-#' @param anno_txt_var_sl ([teal.picks::variables()] or [teal.transform::choices_selected()]) subject level variables to be displayed in the annotation
-#' table, default is `NULL`.
+#' @param anno_txt_var_sl ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' subject level variables to be displayed in the annotation table, default is `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
-#' @param anno_txt_paramcd_rs ([teal.picks::values()] or [teal.transform::choices_selected()]) analysis dataset variables to be displayed in the
-#' annotation table, default is `NULL`.
+#' @param anno_txt_paramcd_rs ([teal.picks::values()] or [teal.transform::choices_selected()])
+#' analysis dataset variables to be displayed in the annotation table, default is `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::values()] instead.
-#' @param facet_var ([teal.picks::variables()] or [teal.transform::choices_selected()]) facet by variable (subject level), `None` corresponds to
-#' `NULL`.
+#' @param facet_var ([teal.picks::variables()] or [teal.transform::choices_selected()])
+#' facet by variable (subject level), `None` corresponds to `NULL`.
 #' `choices_selected` is being deprecated as an argument type and will be removed in the future.
 #' Please use [teal.picks::variables()] instead.
 #' @param ytick_at (`numeric(1)`) bar height axis interval, default is 20
 #' @param href_line (`numeric`) numeric vector to plot horizontal reference lines, default is `NULL`
-#' @param gap_point_val (`numeric(1)`) singular numeric value for adding bar break when some bars are significantly higher
-#' than others, default is `NULL`
+#' @param gap_point_val (`numeric(1)`) singular numeric value for adding bar break when some bars
+#' are significantly higher than others, default is `NULL`
 #' @param show_value (`logical(1)`) boolean of whether value of bar height is shown, default is `TRUE`
 #'
 #' @inherit argument_convention return
@@ -485,11 +485,15 @@ srv_g_waterfall <- function(id,
       ) {
         bar_color_var_selected
       }
-      sort_var_selected <- if (!is.null(sort_var_selected) && sort_var_selected != "None" && sort_var_selected != "") {
+      sort_var_selected <- if (
+        !is.null(sort_var_selected) && sort_var_selected != "None" && sort_var_selected != ""
+      ) {
         sort_var_selected
       }
 
-      facet_var_selected <- if (!is.null(facet_var_selected) && facet_var_selected != "None" && facet_var_selected != "") {
+      facet_var_selected <- if (
+        !is.null(facet_var_selected) && facet_var_selected != "None" && facet_var_selected != ""
+      ) {
         facet_var_selected
       }
 
--- a/R/tm_g_waterfall.R
+++ b/R/tm_g_waterfall.R
@@ -13,15 +13,15 @@
 #' be available in the list passed to the `data` argument of [teal::init()]
 #' @param dataname_rs (`character(1)`) response analysis data used in teal module to label response parameters, needs to
 #' be available in the list passed to the `data` argument of [teal::init()]
-#' @param bar_paramcd (`values` or `choices_selected`) parameter in tumor burden data that will be plotted as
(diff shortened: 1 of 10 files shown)


R/tm_g_waterfall.R
```diff
@@ -2,44 +2,63 @@
...
-#'   Value for breaking high bars.
-#' @param show_value (`logical`)\cr
-#'   Whether to show bar height values.
+#' @param parentname (`character(1)`) parent analysis data used in teal module, usually this refers to `ADSL`,
+#' which is the default.
+#' @param dataname_tr (`character(1)`) tumor burden analysis data used in teal module to plot as bar height, needs to
+#' be available in the list passed to the `data` argument of [teal::init()]
+#' @param dataname_rs (`character(1)`) response analysis data used in teal module to label response parameters, needs to
+#' be available in the list passed to the `data` argument of [teal::init()]
+#' @param bar_paramcd (`values` or `choices_selected`) parameter in tumor burden data that will be plotted as
```
osenan:
```suggestion
#' @param bar_paramcd (`[`teal.picks::values()`] or `[teal.picks::choices_selected`)] parameter in tumor burden data that will be plotted as
```
Let's write it as in the other module arguments that are teal.picks

R/utils_picks.R
```diff
@@ -261,3 +268,79 @@ create_picks_helper <- function(datasets = NULL, x) {
...
+#' Coerce legacy `choices_selected`-based specs to `picks` with deprecation
+#'
+#' @param x (`variables`, `values`, `choices_selected` or `picks`) object.
+#'   A bare [`teal.picks::variables()`] pick is returned unchanged (column selector only; value
+#'   levels follow from data when the pick chain is completed with `create_picks_helper()`).
+#' @param arg_name optional (`character(1)`) argument name.
+#' @param multiple optional (`logical(1)`) whether multiple values are allowed.
+#' If `NULL` (default), it is not validated and inferred from the length of `selected` in the
+#' `choices_selected` object.
+#'
```
osenan, line 284: Please document as well variable `add_values`

R/utils_picks.R
```diff
@@ -261,3 +268,79 @@ create_picks_helper <- function(datasets = NULL, x) {
...
+#' @keywords internal
+#' @noRd
+migrate_value_choices_to_picks <- function(x, # nolint: object_length_linter.
```
osenan, line 287: I was going to say this function is unnecessary as probably is only required in the waterfall module. Probably we could split this into different chunks within the module (check the class, check and force picks arguments, add values, convert `choices_selected` to values and add the variable if it comes from `choices_selected`. On the other hand, because it is used three times, we would add even more code to the module so it is ok to keep it.

R/tm_g_waterfall.R
```diff
@@ -234,234 +317,215 @@ ui_g_waterfall <- function(id,
...
+      #     "Y-axis Interval must be a single positive number."
+      #   ),
+      #   teal::need_input(
+      #     "gap_point_val",
+      #     input$gap_point_val == "" ||
+      #       (!is.na(suppressWarnings(as.numeric(input$gap_point_val))) &&
+      #         checkmate::test_number(suppressWarnings(as.numeric(input$gap_point_val)), lower = 1)),
+      #     "Break High Bars must be a single positive number."
+      #   )
+      # )
```
osenan: I think this is wrong? You have commented out the input validation using shiny::validate, but it should be kept.

R/tm_g_waterfall.R
```diff
@@ -234,234 +317,215 @@ ui_g_waterfall <- function(id,
...
-        .waterfall_picks_selected_var(selectors$facet_var())
-      } else {
-        character()
-      }
-
-      bar_paramcd_one <- bar_paramcd_val[[1L]]
-      validate(
-        need(
-          bar_paramcd_one %in% adtr$PARAMCD,
-          "Tumor burden parameter must be an element of ADTR PARAMCD."
```
osenan: Is this validation required? Or can we remove it?

R/tm_g_waterfall.R
```diff
@@ -234,234 +317,215 @@ ui_g_waterfall <- function(id,
...
-      if (length(add_label_var_sl_name) > 0L && length(add_label_paramcd_rs_val) > 0L) {
-        validate(need(FALSE, "Only one \"Label to Bars\" is allowed."))
-      }
-      if (length(anno_txt_paramcd_rs_val) > 0L) {
-        validate(need(
-          all(anno_txt_paramcd_rs_val %in% adrs$PARAMCD),
-          "Annotation parameters must be elements of ADRS PARAMCD."
-        ))
-      }
-
```
osenan: Again I see here validation that is not present anymore, but we still have the values. If we can I would keep it.

R/tm_g_waterfall.R
```diff
@@ -145,57 +238,47 @@ ui_g_waterfall <- function(id,
...
+      ),
+      tags$div(
+        tags$strong("Sort by"),
+        teal.picks::picks_ui(ns("sort_var"), sort_var),
+        helpText("from ", tags$code(parentname))
+      ),
+      tags$div(
+        tags$strong("Add ADSL Label to Bars"),
+        teal.picks::picks_ui(ns("add_label_var_sl"), add_label_var_sl)
+      ),
```
osenan, line 266: Before migration, this argument could be NULL. The teal developer could set to NULL, the input should be absent from the UI and it shuold not affect the server.

R/tm_g_waterfall.R
```diff
@@ -49,19 +68,57 @@
...
                            bar_var,
-                           bar_color_var = NULL,
+                           bar_color_var,
                            bar_color_opt = NULL,
-                           sort_var = NULL,
-                           add_label_var_sl = NULL,
-                           add_label_paramcd_rs = NULL,
-                           anno_txt_var_sl = NULL,
-                           anno_txt_paramcd_rs = NULL,
-                           facet_var = NULL,
```
osenan: Those arguments were NULL for a reason, that is that argument could be NULL, teal developer could set it to NULL. They should not appear in the UI if NULL and the server function should work fine even if they are set to NULL.
I recommend to keep it here NULL and to modify the UI and check the server.

R/tm_g_waterfall.R
```diff
@@ -71,56 +128,92 @@ tm_g_waterfall <- function(label,
...
+
+  bar_paramcd <- migrate_value_choices_to_picks(bar_paramcd, default_variable_name = "PARAMCD")
+  bar_var <- migrate_choices_selected_to_variables(bar_var)
+  bar_color_var <- migrate_choices_selected_to_variables(bar_color_var)
+  sort_var <- migrate_choices_selected_to_variables(sort_var)
+  anno_txt_var_sl <- migrate_choices_selected_to_variables(anno_txt_var_sl)
+  anno_txt_paramcd_rs <- migrate_value_choices_to_picks(anno_txt_paramcd_rs, default_variable_name = "PARAMCD")
+  facet_var <- migrate_choices_selected_to_variables(facet_var)
+  add_label_var_sl <- migrate_choices_selected_to_variables(add_label_var_sl)
+  add_label_paramcd_rs <- migrate_value_choices_to_picks(add_label_paramcd_rs, default_variable_name = "PARAMCD")
```
osenan, line 167: all of these need `null.ok = TRUE`

osenan (changes requested):
Thanks for the changes but this PR is not good to merge.
The example works and we have improved some parts.

However, there are other changes which I am not sure if they were intentional 100% or not but I think they are not good. The most important changes that cannot be accepted are:
1. There are arguments that were NULL before the migration. This should continue this way, as some inputs are optional. It affects the module assertions, ui and server function. All should be reverted and/or updated in this branch.
2. Input validation has been deleted, we need to keep it and/or update it if required.
3. With a smaller priority, please let's update the documentation as in the other modules. First document the picks then the teal.transform. I know this is not consistent yet across other modules.
4. Finally, let's try to have the CI checks passed as well before the new review, I think the feature branch has check passed.

Finally, consider if working from this branch is helping you or is a constraint that is making the work more difficult. Maybe it is more comfortable to revert the whole module as before the migration and start from scratch. I have seen you have been trying to remove unnecessary utils functions and there might be other artifacts in the server that challenge the work and take time and attention.

R/tm_g_waterfall.R
averissimo: Done in 05d2ddd2

R/tm_g_waterfall.R
averissimo: Not really, the version from `main` didn't have the `NULL`, I've updated it on the feature branch to revert to `main` so the review is easier to compare to the original module ☺️

R/tm_g_waterfall.R
averissimo: Shoot! it is, removed the comments

R/utils_picks.R
averissimo: Done in 135d0805

It is also missing in tmc :-\

R/utils_picks.R
```diff
@@ -261,3 +268,82 @@ create_picks_helper <- function(datasets = NULL, x) {
...
+#' @param default_variable_name optional (`character(1)`) variable name to use if `x` is a bare `values` pick.
+#' @param add_values (`logical(1)`) whether to add a `values` pick if `x` is a bare `variables` pick.
+#' Default is `TRUE`.
+#'
+#' @keywords internal
+#' @noRd
+migrate_value_choices_to_picks <- function(x, # nolint: object_length_linter.
+                                           multiple = NULL,
+                                           arg_name = checkmate::vname(x),
+                                           default_variable_name = NULL,
```
averissimo, line 290: @osenan what do you think about `default_variable_name`?

This actually simplifies the apps a bunch, but defaulting to `PARAMCD` via this argument.

This way we can actually have as argument `mod(..., paramcd = values())`

R/tm_g_waterfall.R
averissimo: They don't, please refer to the version on `main` or the new comparison on this branch as I've reverted the waterfall module on the feature branch ☺️

R/tm_g_waterfall.R
averissimo: Resolving as this was an artifact from the badly implemented module

averissimo: See other comments

R/tm_g_waterfall.R
```diff
@@ -280,91 +341,92 @@ srv_g_waterfall <- function(id,
...
+        ),
+        teal::need_input(
+          "bar_paramcd-values-selected",
+          all(selectors$bar_paramcd()$values$selected %in% data()[[dataname_tr]]$PARAMCD),
+          "All values of Tumor Burden Parameter must be elements of ADTR PARAMCD."
+        ),
+        teal::need_input(
+          "add_label_paramcd_rs-values-selected",
+          all(selectors$add_label_paramcd_rs()$values$selected %in% data()[[dataname_rs]]$PARAMCD),
+          "All values of ADRS Label must be elements of ADRS PARAMCD."
```
averissimo:
```suggestion
          "ADRS Label must be an element of ADRS PARAMCD."
```

R/tm_g_waterfall.R
```diff
@@ -280,91 +341,92 @@ srv_g_waterfall <- function(id,
...
+        ),
+        teal::need_input(
+          c("bar_paramcd-values-selected", "add_label_var_sl-variables-selected"),
+          length(selectors$add_label_var_sl()$variables$selected) == 0 ||
+            length(selectors$add_label_paramcd_rs()$values$selected) == 0,
+          "Only one of 'Add ADSL Label to Bars' and 'Add ADRS Label to Bars' can be selected."
```
averissimo, line 390: This is equivalent to the left side: `"Only one \"Label to Bars\" is allowed"`

R/tm_g_waterfall.R
```diff
@@ -280,91 +341,92 @@ srv_g_waterfall <- function(id,
...
+        ),
+        teal::need_input(
+          "anno_txt_paramcd_rs-values-selected",
+          all(selectors$anno_txt_paramcd_rs()$values$selected %in% data()[[dataname_rs]]$PARAMCD),
+          "All values of Annotation Parameters must be elements of ADRS PARAMCD."
```
averissimo:
```suggestion
          "Annotation Parameters must be elements of ADRS PARAMCD"
```

R/tm_g_waterfall.R
averissimo: This was most probably AI hallucination

R/tm_g_waterfall.R
averissimo: This was most probably AI hallucination

R/tm_g_waterfall.R
osenan: Thanks, I did not know that

R/tm_g_waterfall.R
osenan: Actually I should have checked main. I assumed that the feature branch was respecting main but I should not have taken for granted

R/utils_picks.R
```diff
@@ -84,20 +84,24 @@
 #'
 #' @keywords internal
 #' @noRd
-force_pick_variable_selection <- function(pick, arg_name, multiple = FALSE) {
+force_pick_variable_selection <- function(pick,
```
osenan: maybe we could rename the function, now that it is applied to any pick element not only to `variables`, my suggestions are:
- "force_pick_multiple_argument`
- "force_pick_selection"

This is not a blocker to approve, just if you agree function name can be updated

R/utils_picks.R
osenan: It is actually a very good idea. Let's add an assertion so we ensure it is either NULL or a string with length one as I see a very edge case where user puts two variables with shared values.

tests/testthat/test-tm_g_waterfall.R
```diff
@@ -0,0 +1,252 @@
...
+          classes = "picks_delayed"
+        )
+      },
+      "`bar_paramcd` accepts only a single variable selection"
+    )
+  })
+})
+
+testthat::describe("tm_g_waterfall module creation", {
+  testthat::it("creates a teal module using choices_selected (default method)", {
```
osenan:
```suggestion
  testthat::it("creates a teal module using choices_selected (deprecated option)", {
```

tests/testthat/test-tm_g_waterfall.R
```diff
@@ -0,0 +1,252 @@
...
+      anno_txt_var_sl = anno_txt_var_sl_cs,
+      anno_txt_paramcd_rs = anno_txt_paramcd_rs_cs,
+      facet_var = facet_var_cs,
+      href_line = "-30, 20",
+      plot_height = c(1200, 400, 5000)
+    )
+    testthat::expect_s3_class(mod, "teal_module")
+  })
+
+  testthat::it("creates a teal module using picks (.pick method)", {
```
osenan:
```suggestion
  testthat::it("creates a teal module using picks", {
```

R/tm_g_waterfall.R
```diff
@@ -7,37 +7,58 @@
 #' @inheritParams teal.widgets::standard_layout
 #' @inheritParams teal::module
 #' @inheritParams argument_convention
-#' @param dataname_tr tumor burden analysis data used in teal module to plot as bar height, needs to
+#' @param parentname (`character(1)`) parent analysis data used in teal module, usually this refers to `ADSL`,
```
osenan, line 10: suggestion, shall we add it to argument convention file and just inherit? There are several modules that contain this argument after our migration refactor.

R/tm_g_waterfall.R
osenan: Only `bar_color_opt` [was allowed](https://github.com/insightsengineering/teal.osprey/blob/main/R/tm_g_waterfall.R#L93) to be NULL are we are keeping it.

R/tm_g_waterfall.R
osenan: It is ok

R/tm_g_waterfall.R
```diff
@@ -373,94 +436,96 @@ srv_g_waterfall <- function(id,
...
-      facet_var <- if (!is.null(input$facet_var) && input$facet_var != "None" && input$facet_var != "") {
-        input$facet_var
-      } else {
-        NULL
+
+      facet_var_selected <- if (
+        !is.null(facet_var_selected) && facet_var_selected != "None" && facet_var_selected != ""
+      ) {
+        facet_var_selected
       }
```
osenan: I think you can remove all these lines. If I am not mistaken, if the selected pick variable is not NULL, is not "None" or is not an empty string we are keeping the selected variable. And if it is any of the three exceptions we are keeping the selected variable as well. Just remove it if no reason I cannot see.

R/tm_g_waterfall.R
```diff
@@ -484,16 +549,13 @@ srv_g_waterfall <- function(id,
...
-          qq1,
-          code = bquote({
-            rs_label <- rs_sub %>%
-              dplyr::select(USUBJID, PARAMCD, AVALC) %>%
-              tidyr::pivot_wider(names_from = PARAMCD, values_from = AVALC)
-            anl <- bar_data %>% dplyr::left_join(rs_label, by = c("USUBJID"))
-            anl$USUBJID <- unlist(lapply(strsplit(anl$USUBJID, "-", fixed = TRUE), tail, 1))
-          })
-        )
+        within(qq1, {
```
osenan:
```suggestion
        within(q1, {
```
This seems a typo. There are 3 references to `qq1` in the module. Please fix all of them

R/tm_g_waterfall.R
```diff
@@ -7,37 +7,58 @@
...
-#' annotation table, default is `NULL`
-#' @param facet_var `choices_selected` facet by variable (subject level), `None` corresponds to
-#' `NULL`
-#' @param ytick_at bar height axis interval, default is 20
-#' @param href_line numeric vector to plot horizontal reference lines, default is `NULL`
-#' @param gap_point_val singular numeric value for adding bar break when some bars are significantly higher
-#' than others, default is `NULL`
-#' @param show_value boolean of whether value of bar height is shown, default is `TRUE`
+#' @param sort_var ([`teal.picks::variables`] or [`teal.transform::choices_selected`])
+#' sort by variable (subject level), `None` corresponds to `NULL`.
```
osenan: I do not think this that assigning "None" is anymore the case. The variables need to belong to the dataset. The proper way of not using sort or other variables that are optional is to keep the selected variable `NULL`. I would change this documentation and all the other variables that refer to "None".
Another solution, that implies more work, is to allow teal developer to directly set to `NULL` `sort_bar` and all optional variables. However this means you need to refactor the ui and the server function, in addition to the checks into this module. On the other hand, it is more in concordance with the rest of the modules where sorting, faceting and optional variables can be NULL and can be missing from the UI.
It is up to you I suggest just updating the documentation and remove any reference to "None".

osenan (changes requested):
Hi, many thanks for the changes. I think you did a great work. The module works, the test are improved. All my previous comments have been answered or justified. Apologies for the confusion about the NULL variables.
I have not approved but there are minor comments only to be answered.
1. I suggest to remove all references to None in the documentation and code
2. There are typos and code that can be removed
Other comments please just answer them directly.

R/tm_g_waterfall.R
averissimo: It's not, but I'll rename it to something more meaningful, `q_temp`

R/utils_picks.R
averissimo: `force_pick_selection` seems short and to the point

R/tm_g_waterfall.R
averissimo: Sounds good, do you want to do it here or on a different PR?

R/tm_g_waterfall.R
averissimo: Good point, I've made changes to reflect the `teal.picks` way. So the default is a generic picks with no selection.

This cleans some code in the process

Done in 83503ef4

R/tm_g_waterfall.R
averissimo: Good catch, done!

R/tm_g_waterfall.R
osenan: Can be done on the feature branch, it affects several modules

osenan (approved):
Thank you for the last changes, all good on my side.

PR #349, opened 2026-07-27
340 improve documentation

osenan:
This PR addresses missing points on consistency pattern check for issue #340.
The changes only involve documentation. It verifies that we always document teal.picks/teal.transform objects in the same way (first teal.picks second teal.transform) and that equivalent arguments have similar documentation style.


R/argument_convention.R
```diff
@@ -16,13 +16,17 @@
...
+#' @param arm_var Either a ([`teal.picks::variables()`]) object or a
+#'  ([`teal.transform::choices_selected()`]) object.\cr
+#'  `choices_selected` is being deprecated as an argument type and will be removed in the future.
+#'  Object with all available choices and the pre-selected option for variable
+#'  names that can be used as `arm_var`. Column `arm_var` in the `dataname`
+#'  has to be a factor.
+#'
+#' @param paramcd Either a ([`teal.picks::variables()`]) object or a
+#'  ([`teal.transform::choices_selected()`]) object.\cr
+#'  `choices_selected` is being deprecated as an argument type and will be removed in the future.
```
llrs-roche, line 28: No needed to link as we have just linked it, but I think it is better to make it clear it is a function:

```suggestion
#'  `choices_selected()` is being deprecated as an argument type and will be removed in the future.
```

R/tm_g_ae_oview.R
```diff
@@ -7,13 +7,11 @@
 #' @inheritParams teal.widgets::standard_layout
 #' @inheritParams teal::module
 #' @inheritParams argument_convention
-#' @param flag_var_anl Either a ([`teal.transform::choices_selected`])
-#'   `choices_selected` object or a ([`teal.picks::variables()`])
-#'   object with variables used to count adverse event
+#' @param flag_var_anl Either a ([`teal.picks::variables()`]) object or a
+#'   ([`variables()`]) object.
```
llrs-roche: This is the same. Did you meant choices_selected?

llrs-roche (commented):
Several arguments are documented as:

> Either a ([`teal.picks::variables()`]) object or a ([`variables()`]) object.

Which is the same. Please correct it

R/tm_g_events_term_id.R
```diff
@@ -35,7 +37,7 @@
 #'     tm_g_events_term_id(
 #'       label = "Common AE",
 #'       dataname = "ADAE",
-#'       parent_dataname = "ADSL",
+#'       parentname = "ADSL",
```
llrs-roche, line 40: Won't this break backward compatibility?

R/tm_g_swimlane.R
```diff
@@ -3,41 +3,48 @@
 #' @description
 #'
 #' This is a teal module that generates a `swimlane` plot (bar plot with markers) for `ADaM` data
-#' using [teal.picks::variables()] encodings.
+#' using [variables()] encodings.
```
llrs-roche: Maybe we should mark as code this, but I'm not sure if it is worth to add that this is specific for variables(). I think the initial description should not focus on the implementation details

```suggestion
#' using [`variables()`] encodings.
```

R/tm_g_waterfall.R
```diff
@@ -13,47 +13,47 @@
...
-#' @param bar_var ([`teal.picks::variables`] or [`teal.transform::choices_selected`])
-#' numeric variable from dataset to plot the bar height, e.g., `PCHG`.
-#' `choices_selected` is being deprecated as an argument type and will be removed in the future.
-#' Please use [teal.picks::variables()] instead.
-#' @param bar_color_var ([`teal.picks::variables`] or [`teal.transform::choices_selected`])
-#' color by variable (subject level). Defaults to no selection.
-#' `choices_selected` is being deprecated as an argument type and will be removed in the future.
-#' Please use [teal.picks::variables()] instead.
+#' @param bar_paramcd Either a ([`teal.picks::values()`]) object or a
+#'   ([`variables()`]) object.
```
llrs-roche: Is it true that it admits values? I might have missed that

tests/testthat/test-tm_g_swimlane.R
llrs-roche, line 1: What are these new tests added? Aren't they already on the branch it will be merged? What do they add? I think this comes from fixing some conflicts but this shouldn't be needed on this PR

R/tm_g_waterfall.R
osenan: yes, this argument admits both objects

tests/testthat/test-tm_g_swimlane.R
osenan: I think this comes from fixing conflicts and seems that either the tests are duplicated or the descriptions are not precise, as there are repeated descriptions. I will review and update.

R/tm_g_events_term_id.R
osenan: No, it won't. To my knowledge argument `parentname` or equivalent argument did not existed in any module before migration. Many modules had harcoded `ADSL`  within the module or server function. Explicitly adding `parentname` is an improvement in case teal.developer uses analysis dataset different than "ADSL". And this commit is only using the same argument name across all modules that use `parentname`.

R/tm_g_swimlane.R
osenan: Yes, this is an artifact from the first iteration when we had two methods with S3 dispatch. Nice catch.

R/tm_g_ae_oview.R
osenan: Yes, it is an artifact from changing it with LLM. I should have view that before the PR was ready, sorry!. Fixed!

R/argument_convention.R
osenan: True, changed in all modules

tests/testthat/test-tm_g_swimlane.R
osenan: It actually had duplicated code, tests and wrong descriptions. I updated everything. I think this was coming from branch 340

R/tm_g_events_term_id.R
llrs-roche: Ok, I have verified that it didn't exists before the migration to teal.picks.

tests/testthat/test-tm_g_swimlane.R
```diff
@@ -138,22 +138,26 @@ describe("tm_g_swimlane module creation", {
...
-      ADSL <- rADSL %>%
-        mutate(TRTDURD = as.integer(TRTEDTM - TRTSDTM) + 1) %>%
-        filter(STRATA1 == "A" & ARMCD == "ARM A")
-      ADRS <- rADRS %>%
-        filter(PARAMCD == "LSTASDI" & DCSREAS == "Death") %>%
-        mutate(AVALC = DCSREAS, ADY = EOSDY) %>%
-        rbind(filter(rADRS, PARAMCD == "OVRINV" & AVALC != "NE")) %>%
-        arrange(USUBJID)
-    })
+testthat::describe("test for server module function", {
```
llrs-roche: I removed all `testthat::` prefixes. I am not confident that this PR doesn't miss some updates/code from `redesign_extraction@main`.

tests/testthat/test-tm_g_swimlane.R
llrs-roche: See the comment below, I think there is still something amiss with this branch

llrs-roche (commented):
Good changes, I resolved some comments but I am not sure if this is ready to be merged. Please review the merge and changes from the parent branch carefully

tests/testthat/test-tm_g_swimlane.R
osenan: Ok, I can revert all my changes and replace to what is currently in `test-tm_g_swimlane.R` in the `redesign_extraction@main`

osenan: https://github.com/insightsengineering/teal.osprey/blob/redesign_extraction%40main/tests/testthat/test-tm_g_swimlane.R I can revert to this state, that does not have duplicated tests. And make no changes. I suggest that in testing PR you modify this script the following way:
- split the `it` refering to testServer to a new `describe` block that is exclusive of `testServer`. It can be called "the server function runs without errors" or the name you consider more appropiate.
- The data does not need to be defined in each `it` block. It can be defined at the top of `describe` and reused. Only applies to `testServer`.

llrs-roche: Ok, let's revert the changes here and see if there's need to do changes on tests on a different PR.

llrs-roche (approved):
All looks good now. The CI will fail as tests are failing but we can merge for now as the other PR will address that

PR #350, opened 2026-07-28
Fix tests for redesign

llrs-roche:
# Pull Request

Partially fixes #340 

Address the failing problem for tm_g_patient_profile().

### DIFF
--- a/R/tm_g_patient_profile.R
+++ b/R/tm_g_patient_profile.R
@@ -726,7 +726,6 @@ srv_g_patient_profile <- function(id,
 
       q1 <- if (isTRUE(select_plot()[ae_dataname])) {
         if (all(ADAE$USUBJID %in% ADSL$USUBJID)) {
-
           adae_vars <- unique(c(
             "USUBJID", "STUDYID", "ASTDT",
             "AENDT", "AESOC", "AEDECOD",
@@ -802,7 +801,6 @@ srv_g_patient_profile <- function(id,
 
       q1 <- if (isTRUE(select_plot()[rs_dataname])) {
         if (all(ADRS$USUBJID %in% ADSL$USUBJID)) {
-
           adrs_vars <- unique(c(
             "USUBJID", "STUDYID", "PARAMCD",
             "PARAM", "AVALC", "AVAL", "ADY",
@@ -848,7 +846,6 @@ srv_g_patient_profile <- function(id,
 
       q1 <- if (isTRUE(select_plot()[cm_dataname])) {
         if (all(ADCM$USUBJID %in% ADSL$USUBJID)) {
-
           adcm_vars <- unique(c(
             "USUBJID", "STUDYID", "ASTDT",
             "AENDT", "ASTDT", "CMDECOD",
@@ -903,7 +900,6 @@ srv_g_patient_profile <- function(id,
 
       q1 <- if (isTRUE(select_plot()[ex_dataname])) {
         if (all(ADEX$USUBJID %in% ADSL$USUBJID)) {
-
           adex_vars <- unique(c(
             "USUBJID", "STUDYID", "ASTDT",
             "AENDT", "PARCAT2", "AVAL",
--- a/R/tm_g_patient_profile.R
+++ b/R/tm_g_patient_profile.R
@@ -245,6 +245,11 @@ tm_g_patient_profile <- function(label = "Patient Profile Plot",
     .var.name = "plot_width"
   )
 
+  checkboxes <- c(ex_dataname, ae_dataname, rs_dataname, lb_dataname, cm_dataname)
+  if (sum(!is.na(checkboxes)) < 1L) {
+    stop("Please specify some datanames.")
+  }
+
   args <- as.list(environment())
 
   module(
@@ -459,18 +464,18 @@ srv_g_patient_profile <- function(id,
       })
     }
 
-    # render plot
-    output_q <- shiny::debounce(
+    # Validate inputs
+    validate_q <- shiny::debounce(
       millis = 200,
       r = reactive({
-        obj <- data()
+        obj <- req(data())
+
         teal.reporter::teal_card(obj) <-
           c(
             teal.reporter::teal_card(obj),
             teal.reporter::teal_card("## Module's output(s)")
           )
         obj <- teal.code::eval_code(obj, "library(dplyr)")
-
         patient_id_selected <- selectors$patient_id()$values$selected
         sl_start_date_selected <- pick_selected("sl_start_date", selectors)
         ae_var_selected <- pick_selected("ae_var", selectors)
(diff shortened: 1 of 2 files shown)


tests/testthat/test-tm_g_patient_profile.R
```diff
@@ -37,27 +37,32 @@ describe("tm_g_patient_profile input validation", {
...
-            label = "Patient Profile Plot",
-            patient_id = patient_id_cs,
-            sl_dataname = "ADSL",
-            sl_start_date = sl_start_date_cs
-          ),
-          classes = "picks_delayed"
-        )
-      },
-      "Assertion on 'plot_width' failed"
+    expect_warning(
```
osenan: This test fails. The main problem is that we are trying to implement two tests in a single call, which I think is unnecessary.
If you would like to add a test to see if the warning for deprecation, create a separate test with module arguments that do not produce an error and then expect a warning with deprecation message.

tests/testthat/test-tm_g_patient_profile.R
```diff
@@ -1,31 +1,31 @@
...
-  choices = dplyr::where(is.character),
-  selected = 1L
+  choices = "USUBJID",
+  selected = "USUBJID"
 )
 
 sl_start_date_picks <- teal.picks::variables(
-  choices = dplyr::where(function(x) inherits(x, c("Date", "POSIXct", "POSIXt"))),
-  selected = 1L
+  choices = c("TRTSDTM", "RANDDT"),
```
osenan, line 17: FYI, I tried also to create more generic default for the date input. Unfortunately, it is not enough only to require date input. I think we need to have extra input validation to provide useful information to user about which columns can be selected.
This is not a blocker for this PR but maybe we can create an issue.

tests/testthat/test-tm_g_patient_profile.R
```diff
@@ -103,40 +108,67 @@ describe("tm_g_patient_profile input validation", {
...
+        sl_start_date = sl_start_date_picks,
+        plot_height = c(1200, 400, 5000)
+      ),
+      classes = "picks_delayed"
+    )
     expect_s3_class(mod, "teal_module")
   })
 
-
   data <- within(teal_data(), {
```
osenan, line 161: In my opinion, from here it should be in a separate description block. I would start a new `describe` block representing now we start `testServer` tests. The nature of the test are different (previous ones only call directly the teal module function, from now we test the server function). In addition, as in this commit all tests are within the same description block the test descriptions seem repeated (they are not, but only after taking a look at the test content).

R/tm_g_patient_profile.R
osenan, line 1: Thanks for the implementation of the `validated_q` in this module.
However, you have introduced a bug.

Image
To reproduce it, please try this example:

```
devtools::load_all()
 data <- teal_data() %>%
   within({
     library(nestcolor)
     library(dplyr)
     ADSL <- rADSL
     ADAE <- rADAE %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     ADCM <- rADCM %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     # The step below is to pre-process ADCM to legacy standard
     ADCM <- ADCM %>%
       select(-starts_with("ATC")) %>%
       unique()
     ADRS <- rADRS %>% mutate(ADT = as.Date(ADTM))
     ADEX <- rADEX %>% mutate(ASTDT = as.Date(ASTDTM), AENDT = as.Date(AENDTM))
     ADLB <- rADLB %>% mutate(ADT = as.Date(ADTM), LBSTRESN = as.numeric(LBSTRESC))
   })

 ADSL <- data[["ADSL"]]

 app <- init(
   data = data,
   modules = modules(
     tm_g_patient_profile(
       patient_id = variables(
         choices = USUBJID,
         selected = USUBJID
       ),
        sl_start_date = variables(
          selected = "TRTSDTM",
        ),
       ex_var = variables(
         selected = "PARCAT2",
         choices = "PARCAT2"
       ),
       ae_var = variables(
          selected = "AEDECOD",
        ),
        ae_line_col_var = variables(
          selected = "AESER",
          choices = c("AESER", "AEREL")
        ),
    #    lb_var = NULL,
       sl_dataname = "ADSL",
       ex_dataname = "ADEX",
        ae_dataname = "ADAE",
    #    rs_dataname = "ADRS",
    #    cm_dataname = "ADCM",
#       lb_dataname = "ADLB",
       x_limit = "-28, 750",
     )
   )
 )
 if (interactive()) {
   shinyApp(app$ui, app$server)
 }
```
Now deselect one ADaM domain. In the branch `redesign_extraction@main` if you deselect one domain (with this same example). The plot changes, from two to one plot. In this branch, if you deselect one domain we do not see plot anymore.

If you cannot easily find where is the error, maybe we can revert to organize the `qenv` without `validated_q`. This server module is different, without using `teal.picks::merge_srv` function and having adhoc processing functions depending on the analysis dataset.

osenan (changes requested):
Thanks for fixing the test. I have a suggestion to create an additional describe block.
I cannot approve due to the bug I observe in the module. Consider adding the minimum modifications to the module and just fixing the test if it is difficult fixing the bug, it was working in the reference branch.

tests/testthat/test-tm_g_patient_profile.R
osenan: I think you fixed this now

tests/testthat/test-tm_g_patient_profile.R
llrs-roche: Here I changed the inputs as I think they should match between teal.transform and teal.picks, otherwise we are comparing different things.

R/tm_g_patient_profile.R
llrs-roche: I do see the plot after removing ADAE domain:

image

The issue could be on how plots are rendered. Having said that I think the issue with the test could be solved much easier than this PR.

tests/testthat/test-tm_g_patient_profile.R
llrs-roche: But I wouldn't worry much for this, if we need to increase the test coverage to 80% or 90% then we can deal with these extra input validations needed

tests/testthat/test-tm_g_patient_profile.R
llrs-roche: MMh, in all the other modules I have tested it that way.

But here I changed it on 3f366820

R/tm_g_patient_profile.R
osenan: I repeated and it worked to me, maybe it was some problem in my local configuration

tests/testthat/test-tm_g_patient_profile.R
osenan: You did correctly. I wanted to say that there are problems when the date value is not correct that are not correctly diagnosed either in the console warnings/errors and in the input validation messages. But is not something we should do here.

osenan (approved):
Thanks for the clarifications. We are good to merge. I think it is still good to split the module test in module function and testServer tests.

tests/testthat/test-tm_g_patient_profile.R
osenan: Ok, then I think it is better to keep it consistent rather than to "improve" solely this test script. Apologies, then let's better revert.

PR #351, opened 2026-07-29
Improve code consistency in teal.picks migration

osenan:
It finished what is remaining from issue #340

### Changes:
1. Remove adhoc assignation of attributes and use package util function.
2. Update all module teal.picks defaults that are mandatory and had no default. There is only one exception in patient_profile, as `sl_start_date` argument is mandatory but a default is not useful as it is really data dependant.
3. Additionally, small improvement in tests to remove uninformative warnings.

Question for reviewer:
- Shall I also add tests for all the modified modules so they are called in the test only with non default mandatory arguments, therefore we have more confidence the defaults work?


R/tm_g_patient_profile.R
llrs-roche, line 169: At least one other dataname must be present...

R/tm_g_heat_bygrade.R
```diff
@@ -137,11 +137,27 @@ tm_g_heat_bygrade <- function(
...
+    choices = dplyr::starts_with("AVISIT"),
+    selected = 1L
+  ),
+  ongo_var = teal.picks::variables(
+    choices = dplyr::starts_with("ongo"),
+    selected = 1L
+  ),
+  anno_var = teal.picks::variables(
+    choices = teal.picks::is_categorical(min.len = 2),
+    selected = 2L,
```
llrs-roche, line 154: Why here we select the second one per default?

llrs-roche (commented):
I don't think this is a bad change. I have a couple of commenths though.

However, this requires knowing how the package is used with real datasets. I am not sure if the examples are a subset of what is common, or they include all that is needed. I explored some internal packages and I don't think these defaults prevent them to work.

Regarding testing the defaults. If it is short enough yes, I would add them, something like this could be enough

```r
mod <- suppressWarnings(<module>())
testServer(
      mod$server,
      args = c(list(id = "test_id", data = shiny::reactive(data)), mod$server_args),
      expr = {
        session$setInputs() # Iff needed
        expect_no_error(session$returned())
      }
    )
```

R/tm_g_heat_bygrade.R
osenan: This is based in the example, where we set two choices as selected. I can put as default one value, that is not a  big issue.

osenan: Changed to the first one

R/tm_g_patient_profile.R
osenan: This module is slightly special, in the sense that is different from the rest. My reasons to not set any mandatory analysis dataset as default are the following:

1. In this module, analysis datasets are not independent arguments. They also require that its associated variable is non-NULL. Therefore It also would require that the variable for that analysis dataset is defined.
2. Non mandatory arguments are NULL in other modules, we have to reinforce consistency.
3. It remains clearer for app developer to identify which optional analysis datasets can be defined for the user. It is his or her decision to select which optional analysis datasets will be available. If I set one as default then the module will have one mandatory dataset and one optional that the app developer needs to explicitly set to NULL. But then the rest of the other optional datasets are NULL? Why one optional is non-NULL and the others are? It seems simpler to set all to NULL.
4. The module has enough checks so if app developer tries to set it without at least one optional analysis dataset an informative error message is triggered and the app cannot be loaded.

R/tm_g_heat_bygrade.R
averissimo: I've seen this elsewhere and it's to avoid `STUDYID` or similar type variables, in this case it's probably `USUBJID`

R/tm_g_ae_sub.R
```diff
@@ -53,8 +53,8 @@
 #'
 tm_g_ae_sub <- function(label,
                         dataname,
-                        arm_var,
-                        group_var,
+                        arm_var = teal.picks::variables(dplyr::starts_with("ACTARM")),
+                        group_var = teal.picks::variables(function(x) is.factor(x)),
```
averissimo: Why use a anonymous function, when it can just use `is.factor`?

```suggestion
                        group_var = teal.picks::variables(is.factor),
```

R/tm_g_butterfly.R
```diff
@@ -126,11 +126,14 @@
 tm_g_butterfly <- function(label,
                            dataname,
                            filter_var = NULL,
-                           right_var,
-                           left_var,
-                           category_var,
-                           color_by_var,
-                           count_by_var,
+                           right_var = teal.picks::variables(function(x) is.factor(x)),
+                           left_var = teal.picks::variables(function(x) is.factor(x)),
```
averissimo:
```suggestion
                           right_var = teal.picks::variables(is.factor),
                           left_var = teal.picks::variables(is.factor),
```

R/tm_g_swimlane.R
```diff
@@ -128,7 +128,10 @@
 #'
 tm_g_swimlane <- function(label,
                           dataname,
-                          bar_var,
+                          bar_var = teal.picks::variables(
+                            choices = dplyr::where(is.numeric),
+                            selected = 1L
+                          ),
```
averissimo, line 132: Is it worth the simplication?


```suggestion
                          bar_var = teal.picks::variables(
                            choices = is.numeric,
                            selected = 1L
                          ),
```

tests/testthat/test-tm_g_ae_oview.R
```diff
@@ -75,34 +75,34 @@ describe("tm_g_ae_oview module creation", {
   join_keys(data) <- default_cdisc_join_keys[names(data)]
 
   it("creates a teal module using choices_selected", {
-    mod <- tm_g_ae_oview(
+    mod <- suppressWarnings(tm_g_ae_oview(
```
averissimo, line 78: We really need to silence the warnings via environment variables 💡

R/tm_g_swimlane.R
averissimo: These `ADXX <- data[["ADXX"]] can be removed as they are no longer needed. I think they're the only one left in this package

```suggestion
```

averissimo (approved):
Awesome work here!

Good to go as is. There seems to be some redundant default parameters definition that would be great if you could change.

tests/testthat/test-tm_g_ae_oview.R
osenan: Agree! Let's create an issue

R/tm_g_swimlane.R
osenan: Yes, thank you very much

R/tm_g_swimlane.R
osenan: I did not know that, thanks!

osenan: Changed

PR #355, opened 2026-08-14
AI/manual testing on migration examples

osenan:
Closes issue [#354](https://github.com/insightsengineering/teal.osprey/issues/354)

PR #356, opened 2026-08-17
Fix modules after AI/manual testing

osenan:
Analysis of results of manual AI/manual testing of results revealed 3 modules that required fixes.

While analyzing the AI examples test, I compared the plots and saw differences between the results in the three modules altered in this PR. In the three cases the reason was using a different dataset to perform the calculations.
The modified modules can be tested with their example in the documentation. The prove that the result is now equal is shown in the last three comments on [this PR](https://github.com/insightsengineering/teal.osprey/pull/355).


R/tm_g_heat_bygrade.R
```diff
@@ -479,15 +464,20 @@ srv_g_heat_by_grade <- function(
 
         validate_input(
           c("ae_dataname", "id_var", "visit_var", "heat_var"),
-          all(c(id_var_name, visit_var_name, heat_var_name) %in% colnames(validated_q[["heat_data"]])),
-          "Variables ID, Visit and Heat should be present on heat_data"
+          all(c(id_var_name, visit_var_name, heat_var_name) %in% colnames(ae_dataset)),
```
osenan, line 467: Usually this is forces app developer to modify the teal data so the `ae_dataset` includes the three columns, which is not always the case (in our test and examples in needs to be added manually).

R/tm_g_ae_oview.R
```diff
@@ -165,7 +168,7 @@ tm_g_ae_oview <- function(
     ui = ui_g_ae_oview,
     ui_args = args[names(args) %in% names(formals(ui_g_ae_oview))],
     transformators = transformators,
-    datanames = c("ADSL", dataname)
+    datanames = c(parentname, dataname)
```
averissimo: Should we use here the `.picks_dataname(list(arm_var, flag_var_anl)` instead?

R/tm_g_ae_oview.R
```diff
@@ -365,6 +366,18 @@ srv_g_ae_oview <- function(
...
         )
 
+        validate_input(
+          "arm_var",
+          arm_var_name %in% names(arm_source),
+          "Arm Variable must be present on source dataset."
+        )
+
+        validate_input(
+          "arm_var",
```
averissimo: Why flag here the `arm_var` input? Maybe just use a normal `validate(need(...))` call

R/tm_g_ae_oview.R
averissimo: Those `expression(c(...))` wrappers are redundant

```suggestion
        teal.code::eval_code(q1, bquote(
          plot <- osprey::g_events_term_id(
            term = flags,
            id = ANL$USUBJID,
            arm = ANL[[.(arm_var_name)]],
            arm_N = .(arm_N),
            ref = .(input$arm_ref),
            trt = .(input$arm_trt),
            diff_ci_method = .(input$diff_ci_method),
            conf_level = .(input$conf_level),
            axis_side = .(input$axis),
            fontsize = .(font_size()),
            draw = TRUE
          )
        ))
```

R/tm_g_ae_sub.R
```diff
@@ -122,7 +125,7 @@ tm_g_ae_sub <- function(label,
     ui = ui_g_ae_sub,
     ui_args = args[names(args) %in% names(formals(ui_g_ae_sub))],
     transformators = transformators,
-    datanames = c("ADSL", dataname)
+    datanames = c(parentname, dataname)
```
averissimo: Should we use here the `.picks_dataname(list(arm_var,  group_var)` instead?

R/tm_g_ae_sub.R
```diff
@@ -386,11 +398,11 @@ srv_g_ae_sub <- function(id,
             plot <- osprey::g_ae_sub(
               id = ANL$USUBJID,
               arm = ANL[[arm_var_name]],
-              arm_sl = as.character(ANL[[arm_var_name]]),
+              arm_sl = as.character(parent_data[[arm_var_name]]),
               trt = trt,
               ref = ref,
               subgroups = ANL[group_var_name],
-              subgroups_sl = ANL[group_var_name],
+              subgroups_sl = parent_data[, group_var_name],
```
averissimo, line 406: Why not use `ANL` here? The picks merge should include the variable in the data

Image

R/tm_g_ae_sub.R
```diff
@@ -386,11 +398,11 @@ srv_g_ae_sub <- function(id,
             plot <- osprey::g_ae_sub(
               id = ANL$USUBJID,
               arm = ANL[[arm_var_name]],
-              arm_sl = as.character(ANL[[arm_var_name]]),
+              arm_sl = as.character(parent_data[[arm_var_name]]),
```
averissimo, line 402: Why not use `ANL` here? The picks merge should include the variable in the data

Image

R/tm_g_ae_sub.R
averissimo, line 220: Something seems off with `group_var` picks

Image

R/tm_g_heat_bygrade.R
```diff
@@ -498,6 +488,7 @@ srv_g_heat_by_grade <- function(
             validated_q,
             code = substitute(
```
averissimo: Shouldn't this be using `within()`? 

```diff
diff --git a/R/tm_g_heat_bygrade.R b/R/tm_g_heat_bygrade.R
index 4bacb85d..4f5d2dba 100644
--- a/R/tm_g_heat_bygrade.R
+++ b/R/tm_g_heat_bygrade.R
@@ -484,28 +484,24 @@ srv_g_heat_by_grade <- function(
         teal.reporter::teal_card(validated_q) <- c(teal.reporter::teal_card(validated_q), "### Plot")
 
         if (plot_cm) {
-          validated_q <- teal.code::eval_code(
+          validated_q <- within(
             validated_q,
-            code = substitute(
-              expr = {
-                conmed_data <- dplyr::select(conmed_data_src, dplyr::all_of(unique(c(id_var, visit_var, conmed_var))))
-                conmed_data <- conmed_data %>%
-                  filter(conmed_var_name %in% conmed_level)
-                conmed_data[[conmed_var]] <-
-                  factor(conmed_data[[conmed_var]], levels = unique(conmed_data[[conmed_var]]))
-                formatters::var_labels(conmed_data)[conmed_var] <-
-                  formatters::var_labels(ADCM, fill = FALSE)[conmed_var]
-              },
-              env = list(
-                conmed_data_src = as.name(cm_dataname),
-                ADCM = as.name(cm_dataname),
-                id_var = id_var_name,
-                visit_var = visit_var_name,
-                conmed_var = conmed_var_name,
-                conmed_var_name = as.name(conmed_var_name),
-                conmed_level = input$conmed_level
-              )
-            )
+            {
+              conmed_data <- dplyr::select(conmed_data_src, dplyr::all_of(unique(c(id_var, visit_var, conmed_var))))
+              conmed_data <- conmed_data %>%
+                filter(conmed_var_name %in% conmed_level)
+              conmed_data[[conmed_var]] <-
+                factor(conmed_data[[conmed_var]], levels = unique(conmed_data[[conmed_var]]))
+              formatters::var_labels(conmed_data)[conmed_var] <-
+                formatters::var_labels(ADCM, fill = FALSE)[conmed_var]
+            },
+            conmed_data_src = as.name(cm_dataname),
+            ADCM = as.name(cm_dataname),
+            id_var = id_var_name,
+            visit_var = visit_var_name,
+            conmed_var = conmed_var_name,
+            conmed_var_name = as.name(conmed_var_name),
+            conmed_level = input$conmed_level
           )
         }
```

averissimo (changes requested):
Most changes are small stuff, except a picks in `R/tm_g_ae_sub.R` that is not showing properly

I'll be testing examples a bit more while you fix that ☺️

R/tm_g_ae_sub.R
osenan: Then the calculation is done with a different number of rows compared to main and it returns a different result with the same data. Initially we used `ANL` but the result of the plot should not change due to migration.

image

This screenshot is using ANL, but the correct calculation shuold return 268 patients (like in main)

R/tm_g_ae_sub.R
osenan: Same thing mentioned with tm_g_ae_sub happens here, if we use ANL we use different number of rows to make the plot calculation and we modify the result that is in main, while the data has not change it. We should use the parentname here.

R/tm_g_ae_oview.R
osenan: True, good catch

R/tm_g_ae_oview.R
osenan: yes, changed

R/tm_g_ae_sub.R
osenan: yes, changed

R/tm_g_ae_sub.R
osenan: Fixed, the inputID was wrong

R/tm_g_heat_bygrade.R
osenan: true, changed

R/tm_g_ae_sub.R
osenan: If we use ANL the count is 122 patients, but in main the count is 134 patients. It cannot be that we introduce exactly the same data but after migration result is different.

osenan:
Hi, all the changes have been addressed.
1. Use of `.picks_datanames`
2. Use main `dataname` for variables that were using `parentname`
3. Provided explanation on why to use `parentname` dataset for some plot calculations
4. Fix in UI on `tm_g_ae_oview`
5. Small fixes (validated input, as.expression) and conflicts solved

R/tm_g_ae_sub.R
averissimo: Seems fixed

R/tm_g_ae_sub.R
averissimo: I see that you updated them

averissimo (commented):
Everything seems fine, great job Oriol 💯 

Two things:

1. There are 2 comments in the PR and 
1. Can you just fix the examples I see on the screenshot as they don't use `teal.data::`

Image

osenan:
Thanks, sorry I had answered the comments but they were pending. Now the answers are there. I also added the prefix for `teal.data` in examples, good catch.

R/tm_g_ae_sub.R
averissimo: Awesome explanation and great catch!

R/tm_g_ae_sub.R
averissimo: Brilliant! Thanks

averissimo: It also seems that there are no other occurences of this type of parameter

image

averissimo (approved):
Let's go! 💯

